PR #2, opened 2018-08-16
New GX controls: gx-group and gx-switch

MarkMed:
gx-group:
image

gx-switch:
image

### DIFF
--- a/src/components/group/group.tsx
+++ b/src/components/group/group.tsx
@@ -1,0 +1,31 @@
+import { BaseComponent } from "../common/base-component";
+import { Component, Prop } from "@stencil/core";
+@Component({
+  host: {
+    role: "group"
+  },
+  shadow: false,
+  styleUrl: "group.scss",
+  tag: "gx-group"
+})
+export class Group extends BaseComponent {
+  /**
+   * Attribute that provides the caption to the <leyend> tag
+   */
+  @Prop() caption: string;
+  /**
+   * This attribute defines if the control size will grow automatically,
+   * to adjust to its content size.
+   * If set to `false`, it won't grow automatically and it will show scrollbars
+   * if the content overflows.
+   */
+  @Prop() autoGrow: boolean;
+  render() {
+    return (
+      <fieldset class="form-group">
+        <legend>{this.caption}</legend>
+        <slot />
+      </fieldset>
+    );
+  }
+}
--- a/src/components/group/group.tsx
+++ b/src/components/group/group.tsx
@@ -1,16 +1,13 @@
 import { BaseComponent } from "../common/base-component";
 import { Component, Prop } from "@stencil/core";
 @Component({
-  host: {
-    role: "group"
-  },
   shadow: false,
   styleUrl: "group.scss",
   tag: "gx-group"
 })
 export class Group extends BaseComponent {
   /**
-   * Attribute that provides the caption to the <leyend> tag
+   * Attribute that provides the caption to the <legend> tag
    */
   @Prop() caption: string;
   /**
--- a/src/components/group/readme.md
+++ b/src/components/group/readme.md
@@ -1,0 +1,49 @@
+# gx-group
+
+Web Component where you can group related elements of a form field
+
+## Children
+
+The caption attribute of WC (web component) shall be the caption of <legend> tag.
+
+Other elements inserted from main html document shall be inserted in <slot> tag position.
+
+<!-- Auto Generated Below -->
+
+## Properties
+
+#### autoGrow
+
+boolean
+
+This attribute defines if the control size will grow automatically,
+to adjust to its content size.
+If set to `false`, it won't grow automatically and it will show scrollbars
+if the content overflows.
+
+#### caption
+
+string
+
+Attribute that provides the caption to the <leyend> tag
+
+## Attributes
+
+#### auto-grow
+
+boolean
+
+This attribute defines if the control size will grow automatically,
+to adjust to its content size.
+If set to `false`, it won't grow automatically and it will show scrollbars
+if the content overflows.
+
+#### caption
+
+string
+
+Attribute that provides the caption to the <leyend> tag
+
+---
+
+_Built with [StencilJS](https://stenciljs.com/)_
--- a/src/components/group/readme.md
+++ b/src/components/group/readme.md
@@ -2,12 +2,6 @@
 
 Web Component where you can group related elements of a form field
 
-## Children
-
-The caption attribute of WC (web component) shall be the caption of <legend> tag.
-
-Other elements inserted from main html document shall be inserted in <slot> tag position.
-
 <!-- Auto Generated Below -->
 
 ## Properties
@@ -25,7 +19,7 @@ if the content overflows.
 
 string
 
-Attribute that provides the caption to the <leyend> tag
+Attribute that provides the caption to the <legend> tag
 
 ## Attributes
 
@@ -42,7 +36,7 @@ if the content overflows.
 
 string
 
-Attribute that provides the caption to the <leyend> tag
+Attribute that provides the caption to the <legend> tag
 
 ---
 
--- a/src/components/renders/bootstrap/switch/switch-render.tsx
+++ b/src/components/renders/bootstrap/switch/switch-render.tsx
@@ -26,6 +26,8 @@ export function SwitchRender<T extends Constructor<{}>>(Base: T) {
       }
       // Attributes for input. Values taked form HTML and asigned here
       const inputAttrs = {
+        "aria-checked": this.checked ? "true" : undefined,
+        "aria-disabled": this.disabled ? "true" : undefined,
         checked: this.checked,
         class: "switch",
         disabled: this.disabled,
--- a/src/components/renders/bootstrap/switch/switch-render.tsx
+++ b/src/components/renders/bootstrap/switch/switch-render.tsx
@@ -24,7 +24,7 @@ export function SwitchRender<T extends Constructor<{}>>(Base: T) {
           : `gx-checkbox-auto-id-${autoCheckBoxId++}`;
       }
       const inputAttrs = {
-        "aria-checked": this.checked.toString(),
+        "aria-checked": this.checked ? "true" : undefined,
         "aria-disabled": this.disabled ? "true" : undefined,
         checked: this.checked,
         class: "switch",
--- a/src/components/renders/bootstrap/switch/switch-render.tsx
+++ b/src/components/renders/bootstrap/switch/switch-render.tsx
@@ -4,29 +4,27 @@ export function SwitchRender<T extends Constructor<{}>>(Base: T) {
   return class extends Base {
     caption: string;
     checked: boolean;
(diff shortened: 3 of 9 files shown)


src/components/group/group.tsx
```diff
@@ -0,0 +1,31 @@
+import { BaseComponent } from "../common/base-component";
+import { Component, Prop } from "@stencil/core";
+@Component({
+  host: {
+    role: "group"
```
ncardeli: I'd remove the role attribute as it's redundant with the inner fieldset

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+    private getValueFromEvent(event: UIEvent): boolean {
+      return event.target && (event.target as HTMLInputElement).checked; // Return if HTMLInput is checked or not
+    }
+    handleChange(event: UIEvent) {
+      this.checked = this.getValueFromEvent(event); // Check if input element is checked in each click
+      this.onChange.emit(event);
+    }
+    render() {
+      if (!this.inputId) {
+        // Allways run because is undefined, and !undefined = true
```
ncardeli: I'd remove this comment, I think it's unneeded as the code intention is clear

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+        this.inputId = this.id
+          ? `${this.id}_checkbox`
+          : `gx-checkbox-auto-id-${autoCheckBoxId++}`;
+      }
+      // Attributes for input. Values taked form HTML and asigned here
```
ncardeli: Same as in the previous comment. I think the intention is clear in the code.

src/components/renders/bootstrap/switch/switch.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    let element: HTMLGxSwitchElement;
+    let testWindow;
+    beforeEach(async () => {
+      testWindow = new TestWindow();
+      element = await testWindow.load({
+        components: [Switch],
+        html: "<gx-switch id='id0'>TEST TEXT</gx-switch>"
+      });
+    });
+    //innerHTML Test(?)
```
ncardeli: Using textContent is fine in this case. I'd remove this comment.

src/components/renders/bootstrap/switch/switch.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect(element.textContent.trim()).toEqual("TEST TEXT");
+    });
+    //Caption Test
+    it("should set label caption", async () => {
+      element.caption = "TEST TEXT0";
+      await testWindow.flush();
+      expect(element.querySelector("label").innerHTML).toEqual("TEST TEXT0");
+    });
+    //Checked Test
+    it("should detect if is checked", async () => {
```
ncardeli: Minor typo. Sould say: "should detect if **it** is checked"

src/components/renders/bootstrap/switch/switch.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect(element.querySelector("input").checked).toEqual(true);
+    });
+    //Disabled Test
+    it("should detect if is disabled", async () => {
+      element.disabled = true;
+      await testWindow.flush();
+      expect(element.querySelector("input").disabled).toEqual(true);
+    });
+    //ID Test
+    it("should asign value to 'id' and 'for' attrs", async () => {
```
ncardeli: Typo: assign

src/components/group/readme.md
```diff
@@ -0,0 +1,49 @@
+# gx-group
+
+Web Component where you can group related elements of a form field
+
+## Children
```
ncardeli: We use the Children section to explain if the children should have certain structure. Check <gx-card> for example. In this case, the web component accepts any kind of children, so I would remove the section entirely.

src/components/group/group.tsx
```diff
@@ -0,0 +1,31 @@
...
+  },
+  shadow: false,
+  styleUrl: "group.scss",
+  tag: "gx-group"
+})
+export class Group extends BaseComponent {
+  /**
+   * Attribute that provides the caption to the <leyend> tag
```
ncardeli: Typo: legend

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
+import { EventEmitter } from "@stencil/core";
+type Constructor<T> = new (...args: any[]) => T;
+export function SwitchRender<T extends Constructor<{}>>(Base: T) {
+  return class extends Base {
+    caption: string;
+    checked: boolean;
+    disabled = false; // This takes its value form HTML
```
ncardeli: Not sure about this comment, I'd remove it.

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+    element: HTMLElement;
+    id: string;
+    private inputId: string;
+
+    onChange: EventEmitter;
+    private getValueFromEvent(event: UIEvent): boolean {
+      return event.target && (event.target as HTMLInputElement).checked; // Return if HTMLInput is checked or not
```
ncardeli: I'd remove this comment, I think it's unneeded as the code intention is clear

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+    handleChange(event: UIEvent) {
+      this.checked = this.getValueFromEvent(event); // Check if input element is checked in each click
```
ncardeli: I'd remove this comment, I think it's unneeded as the code intention is clear

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+    render() {
+      if (!this.inputId) {
+        // Allways run because is undefined, and !undefined = true
+        this.inputId = this.id
+          ? `${this.id}_checkbox`
+          : `gx-checkbox-auto-id-${autoCheckBoxId++}`;
+      }
+      // Attributes for input. Values taked form HTML and asigned here
+      const inputAttrs = {
+        "aria-checked": this.checked ? "true" : undefined,
```
ncardeli: I'd use a more compact expression to achieve the same result:

`"aria-checked": this.checked.toString(),`

src/components/renders/bootstrap/switch/switch-render.tsx
```diff
@@ -0,0 +1,50 @@
...
+        "aria-disabled": this.disabled ? "true" : undefined,
+        checked: this.checked,
+        class: "switch",
+        disabled: this.disabled,
+        id: this.inputId,
+        onChange: this.handleChange.bind(this),
+        type: "checkbox"
+      };
+      return (
+        <div class="form-group">
```
ncardeli: The `<div>` element is unneeded. You can remove it and the end result will be the same. It will be added when used inside a gx-form-field element.

src/components/switch/readme.md
```diff
@@ -0,0 +1,75 @@
+# gx-switch
+
+An input element for switch/toggle options.
```
ncardeli: I'd change this for something like:

A toggle switch control that enables to select between two options.

src/components/switch/readme.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Children
```
ncardeli: I'd remove this section. See the review comment for the readme.md file of gx-group, please.

src/components/switch/switch.scss
```diff
@@ -0,0 +1,107 @@
...
+          color: #868e96;
+          cursor: not-allowed;
+          &::before {
+            background-color: #e9ecef;
+          }
+        }
+      }
+    }
+  }
+  .switch.switch-sm {
```
ncardeli: These set of rules should be written using SASS

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+  host: {
+    role: "switch"
+  },
+  shadow: false,
+  styleUrl: "switch.scss",
+  tag: "gx-switch"
+})
+export class Switch extends SwitchRender(BaseComponent) {
+  /**
+   * Attribute that provides the caption to the <label> tag.
```
ncardeli: I woudln't mention the `<label>` here. The usage of a `<label>` element is an implementation detail that should be hidden to the component end user.

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  @Prop() caption: string;
+  /**
+   * Indicates that the checked is selected by default.
```
ncardeli: I'd change this for something like:

Indicates if the switch is checked by default.

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  @Prop({ mutable: true })
+  checked: boolean;
+  /**
```
ncardeli: Take a look at the `disabled` property of `<gx-checkbox>`. I'd copy/paste the comment here.

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+   * This attribute lets specify if the element is disabled.
+   */
+  @Prop() disabled = false;
+  /**
+   * The 'change' event is emitted when a change to the element's value is committed by the user.
+   */
+  @Event() onChange: EventEmitter;
+
+  /**
```
ncardeli: Take a look at the `id` property of `<gx-checkbox>`. I'd copy/paste the comment here.

src/components/renders/bootstrap/switch/switch-render.tsx
MarkMed: Cannot use the ```.toString()``` method because it returns the error below:
```
Cannot read property 'toString' of undefined
```

src/components/switch/switch.tsx
MarkMed: The comment looks very similar to ```<gx-checkbox>``` control. Also explains how control ```id``` value works through all the control. Explains wht happens if ou assign an ```id``` to the control or not.
Shall I remove that? i think that comment is useful, don't u think?

MarkMed: The comment looks very similar to ```<gx-checkbox>``` control. But also explains how control ```id``` value works through all the control. Explains wht happens if u assign an ```id``` to the control or not.
Shall I remove that? i think that comment is useful, don't u think?

ncardeli: I think the fact that there's an input and a label element is hidden to the user that reads this doc, so there's no much value explaining that and could lead to confusion.

MarkMed: Got it!

PR #4, opened 2019-02-11
Fix to form-field causing error when no label was included.

ag-silva:


src/components/renders/bootstrap/form-field/form-field-render.tsx
```diff
@@ -63,7 +63,8 @@ export class FormFieldRender implements IRenderer {
       if (nativeInput) {
         nativeInput.setAttribute("data-part", "field");
       }
-      if (nativeInputId) {
+      const innerLabel: any = formField.element.querySelector("label");
+      if (nativeInputId && innerLabel) {
         formField.element
```
ncardeli:
```suggestion
         innerLabel
```

src/components/renders/bootstrap/form-field/form-field-render.tsx
```diff
@@ -63,7 +63,8 @@ export class FormFieldRender implements IRenderer {
...
           .querySelector("label")
```
ncardeli:
```suggestion
```

PR #9, opened 2019-03-07
 New Chronometer WebComponent for GX Chronometer Control Type.

ggallotti:


ggallotti:
@ncardeli Could you review it? Had some unintended  formatting changes (package.json and package-lock.json) due to Prettier after merging from master as one file (components.d.ts) had a merge error.

src/components/chronometer/chronometer.tsx
```diff
@@ -0,0 +1,204 @@
...
+      case TimerState.Stopped:
+        this.stop();
+        break;
+      case TimerState.Reset:
+        this.reset();
+        break;
+      default:
+        break;
+    }
+  }
```
ncardeli:
```suggestion
  }

```

src/components/chronometer/chronometer.tsx
```diff
@@ -0,0 +1,204 @@
...
+  @Method()
+  reset() {
+    this.stop();
+    this.value = 0;
+    this.startedTime = 0;
+    this.elapsedTime = 0;
+    this.handleChange();
+  }
+
+  handleChange() {
```
ncardeli: Consider calling this function using the Watch decorator on the value property, instead of manually calling it after changing the value property.

src/components/chronometer/readme.md
```diff
@@ -0,0 +1,63 @@
+# gx-chronometer
+
```
ncardeli: Please add a brief description of what the component does and a simple example.

src/components/chronometer/chronometer-timer-state.tsx
```diff
@@ -0,0 +1,5 @@
+export enum TimerState {
```
ncardeli: I'd move this enum to chronometer.tsx
I think it doesn't add much to keep it in it's own file.

src/components/chronometer/chronometer.e2e.ts
```diff
@@ -0,0 +1,48 @@
...
+    await page.setContent(
+      "<gx-chronometer unit='1' value='3000' interval='1500' max-value-text='Final' max-value='9000'></gx-chronometer>"
+    );
+    element = await page.find("gx-chronometer");
+  });
+
+  const delay = value =>
+    new Promise(resolve => setTimeout(() => resolve(), value));
+
+  it("Chronometer not started", async () => {
```
ncardeli: I'd change the tests description. The idea with the `it` function is to describe the expected behavior.
In this case I'd use "Sets the chronometer state to started" or "Starts the chronometer".

src/components/chronometer/chronometer.e2e.ts
```diff
@@ -0,0 +1,48 @@
+import { E2EElement, E2EPage, newE2EPage } from "@stencil/core/testing";
+
+describe("gx-chronometer", () => {
```
ncardeli: I'd like to have some tests where the chronometer is started/stopped/reset by changing the state property, instead of calling the methods.

src/components/chronometer/chronometer-timer-state.tsx
ggallotti: Stencil discourage having 2 exports in the same file**. If set as private, then components.d.ts would not compile. 

> @Component() should be the only export of the module. Numerous export statements within a component module
>            may cause undesirable bundling output, leading to unoptimized lazy loading. Create a new auxiliary `.ts`
>            file in order to export shared functionality.

PR #16, opened 2019-03-28
gx-gauge

MarkMed:
New gx-gauge


src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+  private calcThickness() {
+    return typeof this.thickness === "number" &&
+      (this.thickness > 0 && this.thickness <= 100)
+      ? this.thickness / 5
+      : 10;
+  }
+
+  ////////////////////////////////////////
+
+  render() {
```
ncardeli: Please take a second look at the `render` method. It got too big. Maybe it would be a good idea to split it in a `renderCircle` and a `renderLine` method.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+      let acumul = 0;
+      //////////////////////////////////////////
+      function calcPositionRange(preValue, value) {
+        acumul += preValue;
+        return value + acumul;
+      }
+      //////////////////////////////////////////
+      for (let i = childRanges.length - 1; i >= 0; i--) {
+        svgRanges.push(
+          <circle
```
ncardeli: The logic that creates a `<circle>` element could be extracted to a new function to help make the code more compact and readable.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+    const childRanges = Array.from(
+      this.element.querySelectorAll("gx-gauge-range")
+    );
+    this.maxValue = this.totValues + this.minValue;
+    this.totValues = this.maxValue - this.minValue;
+    if (this.gaugeType === "Circle") {
+      const svgRanges = [];
+      let acumul = 0;
+      //////////////////////////////////////////
+      function calcPositionRange(preValue, value) {
```
ncardeli: I'd extract this function out of this method. I'd put it outside the class.
The way it is defined now, it is getting defined every time the render method is callled.
Additionally, the `acumul` variable is updated by the funciton, but it's not used anywhere.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
+import {
```
ncardeli: Please take a look at some "magic" numbers in the code. I'd like these number to be constants, so be know the semantics behind the number.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+  @State() children = [];
+
+  @State() minorSize: number;
+
+  @Listen("gxGaugeRangeDidLoad")
+  onGaugeRangeDidLoad({ detail: childRange }) {
+    this.children = [...this.children, childRange];
+    this.totValues += childRange.amount;
+    childRange.element.addEventListener("gxGaugeRangeDidUnload", () => {
+      const inDex = this.children.findIndex(x => x === childRange);
```
ncardeli: Sorry for nitpicking here, but out of curiosity, why the capital D in `inDex`?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+                "px"
+            }}
+          />
+          {this.showValue ? (
+            <span
+              class="marker"
+              style={{
+                display: this.showValue ? "" : "none",
+                height:
+                  this.minorSize * 0.795 -
```
ncardeli: Please consider using [string literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in cases like this.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+            {this.showValue ? (
+              <div
+                style={{
+                  color: "" + this.styleCenterTextColor,
+                  "font-size":
+                    (this.minorSize * 0.795 -
+                      this.calcThickness() / 2 * (this.minorSize / 100)) /
+                      8 +
+                    "px",
+                  "mix-blend-mode": "difference"
```
ncardeli: I'd put this fixed value in the `scss` file

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+          </div>
+        </div>
+      );
+    } else if (this.gaugeType === "Line") {
+      const divRanges = [];
+      const divRangesName = [];
+      let currentMargin = 0;
+      let currentMargin2 = 0;
+      //////////////////////////////////////////
+      function calcPositionRange(preValue) {
```
ncardeli: Not sure if it justifies having a function to increment a counter. Maybe just use `currentMargin += preValue;`

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+          class="gaugeContainerLine"
+          style={{
+            border: this.styleBorder
+              ? this.styleBorderWidth + "px solid " + this.styleBorderColor
+              : "",
+            height: 5 * this.calcThickness() + "px"
+          }}
+        >
+          <div class="rangesContainer">
+            {divRanges.map(gaugeRange => gaugeRange)}
```
ncardeli: This would be the same as `{divRanges}`, right?
There's no need for the `map` call.

```suggestion
            {divRanges}
```

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+          </div>
+          <div class="namesContainer">
+            {divRangesName.map(gaugeRange => gaugeRange)}
```
ncardeli:
```suggestion
            {divRangesName}
```

src/components/map/map.tsx
```diff
@@ -94,6 +100,10 @@ export class Map implements IComponent {
       {}
     ).addTo(this.map);
     this.gxMapDidLoad.emit(this);
+    this.map.addEventListener("click", ev => {
+      const clickCoords = ev.latlng;
```
ncardeli:
```suggestion
```

src/components/map/map.tsx
```diff
@@ -94,6 +100,10 @@ export class Map implements IComponent {
...
+      this.mapClick.emit(clickCoords);
```
ncardeli:
```suggestion
      this.mapClick.emit(ev.latlng);
```

src/components/gauge-range/readme.md
```diff
@@ -0,0 +1,24 @@
+# gx-range
+
+Component using within `gx-gauge` to draw and show its information.
```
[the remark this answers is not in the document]
MarkMed: I am talking about `gx-gauge-range` component. Using that component you can draw and show its information _(its name)_ within `gx-gauge`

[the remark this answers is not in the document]
ncardeli: Yes, I'm aware, but as far as I understand, `gx-gauge-range` components are used to define each range, right? I suggested the change to be more specific about what the component does in the context of `gx-gauge`.

src/components/gauge/gauge.tsx
MarkMed: To call it like _index_ and prevent use reserved word

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+   * Value in *hex, rgb, rgba, hsl, cmyk* format or *color name*.
+   * _(Gray by Default)_
+   */
+  @Prop() styleCenterTextColor = "gray";
+
+  /**
+   * Property of type Style.
+   * Define if shadow will display or not. Default is disabled.
+   */
+  @Prop() styleShadow = false;
```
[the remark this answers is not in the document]
MarkMed: I was looking to pass this property to SASS style but it must stay in the `.tsx` file because in the JSX I must evaluate another variable.

src/components/gauge/gauge.tsx
MarkMed: I did not know that! 😲

src/components/gauge/gauge.tsx
MarkMed: Those numbers are to take a percentage of a variable number. I do not know how I can explain the purpose of the numbers by assigning them a name.
Perhaps `const gaugeContainer80percent = 0.8;`?
It is really necessary to explain each number? It's just used in a formula to calculates the size to be responsive.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+  /**
+   * This allows specifying the width of the circumference _(When gauge is Circle type)_ and the width of the bar _(When gauge is Line type)_ in % relative the component size.
+   *
+   */
+  @Prop() thickness = 10;
+
+  @State() maxValue: number;
+
+  @State() totValues = this.minValue;
```
ncardeli: I'd rather have a more descriptive name, regardless if its a bit longer, like `totalValues`, or even maybe just `total`.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+      this.children.splice(index, 1, childRange);
+      this.totValues = 0;
+      for (const childInstance of this.children) {
+        this.totValues += childInstance.amount;
+      }
+    });
+  }
+
+  componentWillLoad() {
+    this.totValues = 0;
```
ncardeli: Please verify that this won't fire an additional call to the render method. Maybe setting this value in the property declaration?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+      (this.currentValue - this.minValue) *
+      100 /
+      (this.maxValue - this.minValue)
+    );
+  }
+
+  private renderCircle(childRanges) {
+    const svgRanges = [];
+    let acumulation = 0;
+    //////////////////////////////////////////
```
ncardeli: I feel this doesn't add much to the code, I'd remove it.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+    function calcPositionRange(preValue, value) {
```
ncardeli: Does this need a separate function? It's called only from one place to do a simple operation.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+      acumulation += preValue;
+      return value + acumulation;
+    }
+
+    function addSVGCircle(currentChild, nextChild, component) {
```
ncardeli: How about `renderSvgCircle` instead? This function isn't really adding anything, right?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+          amount={currentChild.amount}
+          stroke-width={`${component.calcThickness()}%`}
+        />
+      );
+    }
+    //////////////////////////////////////////
+    for (let i = childRanges.length - 1; i >= 0; i--) {
+      svgRanges.push(addSVGCircle(childRanges[i], childRanges[i + 1], this));
+    }
+    svgRanges.reverse();
```
ncardeli: Maybe instead of reversing this array, we should just add the items to the beginning of the array?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+    return (
+      <div
+        class="svgContainer"
+        style={{
+          height: `${this.minorSize * 1}px`,
```
ncardeli: Why do we need to multiply by 1?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+   */
+  @Prop() thickness = 10;
+
+  @State() maxValue: number;
+
+  @State() totValues = this.minValue;
+
+  @State() children = [];
+
+  @State() minorSize: number;
```
ncardeli: Would this be the minimum size? Maybe `minimumSize` would be a better name.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+      return (
+        <circle
+          r="39.59%"
+          cx="50%"
+          cy="50%"
+          stroke={currentChild.color}
+          stroke-dasharray={`${2.488 /
+            (component.totValues / 100) *
+            calcPositionRange(
+              !!nextChild ? parseInt(nextChild.getAttribute("amount"), 10) : 0,
```
ncardeli: If you are checking if nextChild is a truthy value, you don't need to double negate it.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+        >
+          {this.showValue ? (
+            <div
+              style={{
+                "font-size": `${(this.minorSize * 0.795 -
+                  this.calcThickness() / 2 * (this.minorSize / 100)) /
+                  8}px`
+              }}
+            >
+              {this.calcPercentage() > 100
```
ncardeli: Consider this expression instead:

```suggestion
              {Math.min(Math.max(Math.round(this.calcPercentage()), 0), 100)}
```

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+      </div>
+    );
+  }
+
+  private renderLine(childRanges) {
+    const divRanges = [];
+    const divRangesName = [];
+    let currentMargin = 0;
+    //////////////////////////////////////////
+    function calcPositionRange(preValue) {
```
ncardeli: Not sure of having a function for this.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+              component.totValues}%`
+          }}
+        >
+          {currentChild.name}
+        </span>
+      );
+    }
+    //////////////////////////////////////////
+    for (let i = childRanges.length - 1; i >= 0; i--) {
+      // create a function to return this structure (like addSVGCircle)
```
ncardeli: I think you forgot to remove this comment.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+            width: `${parseInt(currentChild.getAttribute("amount"), 10) *
+              100 /
+              component.totValues}%`
+          }}
+        >
+          {currentChild.name}
+        </span>
+      );
+    }
+    //////////////////////////////////////////
```
ncardeli: Please reconsider using this type of comments. They don't really add much for other developers that are reading this code.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+    for (let i = childRanges.length - 1; i >= 0; i--) {
+      // create a function to return this structure (like addSVGCircle)
+      divRanges.push(addLineRanges(childRanges[i], childRanges[i + 1], this));
+      divRangesName.push(addRangeCaption(childRanges[i], this));
+    }
+    divRanges.reverse();
```
ncardeli: Plase, consider adding at the beginning instead of adding to the end and then reversing.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+   * Property of type Style.
+   * Define if shadow will display or not. Default is disabled.
+   */
+  @Prop() styleShadow = false;
+
+  /**
+   * This property allows you to select the gauge type. _(Circle or Line)_.
+   * Default is linear type.
+   */
+  @Prop() gaugeType: "line" | "circle" = "line";
```
ncardeli: How about if we use just `type`?

src/components/gauge/readme.md
```diff
@@ -0,0 +1,37 @@
+# gx-gauge
+
+This component allows you to display information in the form of ranges, and decide whether you want to show it using linear or circular gauges.
```
ncardeli:
```suggestion
This component allows displaying information in the form of ranges, and decide whether you want to show it using linear or circular gauges.
```

src/components/gauge/readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+Use `gx-gauge-range` element to set the number of ranges and the max value the gauge will have.
+
```
ncardeli: Adding a sample would be really nice.

src/components/map/map.tsx
```diff
@@ -50,10 +50,16 @@ export class Map implements IComponent {
    */
```
ncardeli: Let's remove this commit from this PR so we don't mix the new gauge control with a fix on the map control.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+
+describe("gx-gauge", () => {
+  let element: E2EElement;
+  let page: E2EPage;
+
+  beforeEach(async () => {
+    page = await newE2EPage();
+  });
+
+  it("should be drawn w/o attrs", async () => {
```
ncardeli:
```suggestion
  it("should be drawn without attributes", async () => {
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.setContent("<gx-gauge></gx-gauge>");
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    expect(element).toBeTruthy();
+    expect(await element.find(".gaugeContainerLine")).toBeTruthy();
+  });
+  /////////////////// Type test ///////////////////////
+
+  it("should render with the line type", async () => {
+    await page.setContent("<gx-gauge gauge-type='Circle'></gx-gauge>");
```
ncardeli: Shouln't we just test for `line` type?

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.waitForChanges();
+    expect(await element.find(".gaugeContainerLine")).toBeTruthy();
+  });
+
+  it("should render with the circle type", async () => {
+    await page.setContent("<gx-gauge gauge-type='Circle'></gx-gauge>");
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
```
ncardeli: Shouldn't we expect something else to ensure that a circle gauge was drawn?

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+  /////////////////// Values test ///////////////////////
+  it("should set min value", async () => {
+    await page.setContent(
+      "<gx-gauge gauge-type='Circle' min-value='0'></gx-gauge>"
+    );
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+    expect(await element.getProperty("minValue")).toEqual(0);
```
ncardeli: I'd like to see some expectations on the effect of setting the `minValue` other than if the value is set in the property.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+  it("should render with the circle type", async () => {
+    await page.setContent("<gx-gauge gauge-type='Circle'></gx-gauge>");
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+  });
+
+  /////////////////// Values test ///////////////////////
+  it("should set min value", async () => {
```
ncardeli:
```suggestion
  it("should set the minimum value", async () => {
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+      "<gx-gauge gauge-type='Circle' min-value='0'></gx-gauge>"
+    );
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+    expect(await element.getProperty("minValue")).toEqual(0);
+  });
+
+  it("should set current value", async () => {
```
ncardeli:
```suggestion
  it("should set the current value", async () => {
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.setContent(
+      "<gx-gauge gauge-type='Circle' min-value='0' current-value='25'></gx-gauge>"
+    );
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+    expect(await element.getProperty("minValue")).toEqual(0);
+    expect(await element.getProperty("currentValue")).toEqual(25);
```
ncardeli: I'd like to see other expectations tested too. How the rendered HTML is affected by this properties?

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+  });
+
+  it("should show the current value", async () => {
+    await page.setContent(
+      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet' class='hydrated'></gx-gauge-range></gx-gauge>"
```
ncardeli: Please, consider using string literals to write this HTML in a more readable way.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.setContent(
+      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true' style-shadow='false'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet' class='hydrated'></gx-gauge-range></gx-gauge>"
+    );
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    await page.waitForChanges();
+    expect(await element.find(".gaugeContainerLine")).toBeTruthy();
+    expect(await element.find("span.marker")).toBeTruthy();
+    expect(await element.find(".minMaxDisplay")).toBeTruthy();
```
ncardeli: I would like to see an expectation for the `Violet` text.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+    expect(await element.getProperty("thickness")).toEqual(20);
+  });
+
+  it("should set a marker in Line", async () => {
+    await page.setContent(
+      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true' style-shadow='false'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet' class='hydrated'></gx-gauge-range></gx-gauge>"
```
ncardeli:
```suggestion
      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true' style-shadow='false'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet'></gx-gauge-range></gx-gauge>"
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    );
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await page.waitForChanges();
+    await page.waitForChanges();
+    expect(await element.find("svg")).toBeTruthy();
+    expect(await element.find("span.marker")).toBeTruthy();
+  });
+
+  it("should set shadows", async () => {
```
ncardeli:
```suggestion
  it("should show shadows", async () => {
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    const gauge = await element.find("div.gauge");
+    await page.waitForChanges();
+    expect(await element.find(".gaugeContainerLine")).toBeTruthy();
+    expect(await gauge.getAttribute("style")).toEqual("box-shadow: none;");
+  });
+
+  /////////////////// gx-range tests ///////////////////////
+  it("should correctly draw the range in Line", async () => {
+    await page.setContent(
+      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true' style-shadow='false' thickness='50'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet' class='hydrated'></gx-gauge-range></gx-gauge>"
```
ncardeli:
```suggestion
      "<gx-gauge gauge-type='Line' min-value='0' current-value='25' show-value='true' style-shadow='false' thickness='50'><gx-gauge-range color='rgba(116, 16, 216, 1)' amount='100' name='Violet'></gx-gauge-range></gx-gauge>"
```

src/components/gauge/gauge.tsx
MarkMed: 😅

src/components/gauge/gauge.tsx
MarkMed: 🤔 😲

src/components/gauge/gauge.tsx
MarkMed: I make a double negate to obtain true or false according to if exist or not the nextChild.

`
if nextChild exist => true => true option of ternary
`
`
if nextChild is undefined => false => false option of ternary
`

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+  @Prop() showValue = false;
+
+  /**
+   * The Minimum Value of the gauge
+   *
+   */
+  @Prop() minValue: number;
+
+  /**
+   * The Currecnt value indicanding in the gauge
```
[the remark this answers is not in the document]
MarkMed: I think that we must make emphasis on the kind of value that we refer to.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+   *
+   */
+  @Prop() currentValue: number = this.minValue;
```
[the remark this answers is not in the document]
MarkMed: I called it that way because I think we should differentiate the values ​​and assign them a role. And in this way, there will not chance to confuse or wonder for the purpose of the property.
Likewise, Do you think that it should be called `value`?

src/components/gauge/gauge.tsx
MarkMed: Nope. The `componentWillLoad` is called just after the component updates and render. And at that moment, the state `totValues` (now `totalValues`) must be restarted with 0 as value.
In short, `totalValues` will only be restarted before the component load.

src/components/gauge/gauge.tsx
MarkMed: Yeap, I think that it's necessary to. Because including the function logic in the place where it's called, the code will be hard to read. I think that is better in that way, separating the code for better understanding.

src/components/gauge/gauge.tsx
MarkMed: 🤨
The function... add a SVG circle to a list of circles. Each part of the gauge (the ranges) is a `<circle>`.
When `renderCircle` is called by render method, it takes the number of ranges (`<gx-gauge-range>`)  and call the `addSVGCircle` to add circles. The `addSVGCircle` only do that, add svgs circles to a list, it does not render anything.

src/components/gauge/gauge.tsx
MarkMed: I made it in that way for take the order of HTML and draw the ranges.
It's like... _"The last range added is the first range in the gauge"_
image

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+              (this.calcThickness() * (this.minorSize / 100) -
+                this.minorSize / 100)}px`
+          }}
+        />
+        {this.showValue ? (
+          <span
+            class="marker"
+            style={{
+              display: this.showValue ? "" : "none",
+              height: `${this.minorSize * 0.795 -
```
ggallotti: Why 0.795?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+        )}
+        <div
+          class="gauge"
+          style={{
+            "box-shadow":
+              (this.minorSize <= 300 && this.thickness <= 25) ||
+              !this.styleShadow
+                ? "none"
+                : "",
+            height: `${this.minorSize * 0.792 -
```
ggallotti: 0.792? Is this a magic number ?

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(element).toBeTruthy();
+    expect(await element.find(".gaugeContainerLine")).toBeTruthy();
+  });
+  /////////////////// Type test ///////////////////////
+
+  it("should render with the line type", async () => {
+    await page.setContent("<gx-gauge gauge-type='Circle'></gx-gauge>");
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    await element.setAttribute("gauge-type", "Line");
```
ggallotti: shouldn't be 'line' in lowercase here?

src/components/gauge/gauge.tsx
ggallotti: +1

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+              this.calcThickness() * (this.minorSize / 100)}px`,
+            width: `${this.minorSize * 0.792 -
+              this.calcThickness() * (this.minorSize / 100)}px`
+          }}
+        >
+          {this.showValue ? (
+            <div
+              style={{
+                "font-size": `${(this.minorSize * 0.795 -
```
ggallotti: Same here

src/components/gauge/gauge.tsx
ncardeli: `addSVGCircle` isn't really adding circles to a list, it just returns circles (that its caller might add to a list), right?
That's why I suggest renaming it.

src/components/gauge/gauge.tsx
ncardeli: I understand, but what I'm suggesting is that you can achieve the same result without the double negation.

src/components/gauge/gauge.tsx
ncardeli: I'm just suggesting an alternative way of having the items added to the array so you don't have to reverse it.

MarkMed: I was looking for another way to draw each `<circle>` of SVG. But I need to first reverse the list of  `<gx-gauge-range>`  that are within  `<gx-gauge>`, then passing one by one and its next for calculate the `stroke-dasharray`. And then reverse the list of `<circle>`s to add all `<circle>`s  within the SVG (with the `stroke-dasharray` already calculated)

MarkMed:
*Reverting changes done in map and previously committed by mistake

src/components/gauge/gauge.tsx
ncardeli: @MarkMed Did you consider this suggestion?

MarkMed: Ready! 😃

src/components/gauge/gauge.tsx
MarkMed: Changes done! commit 3c4d33abc5cf0d9f0a62ddbde15cbf38f1846857

src/components/gauge/gauge.tsx
MarkMed: Yeeeap! I do dat with the splice method. Check the last commit 3c4d33abc5cf0d9f0a62ddbde15cbf38f1846857

src/components/gauge/readme.md
MarkMed: Ready! Fix it in commit  fd99fb8dc921c4e781ec9f22a6daac7000004d47

src/components/gauge-range/gauge-range.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Color property defines the color of range background.
+   * Color value can be in any color model *(hex, rgb, rgba, hsl, cmyk)*.
+   *
+   */
+  @Prop() color: string;
+
+  ////////////////////////////////////////
+
+  @Event() gxGaugeRangeDidLoad: EventEmitter;
```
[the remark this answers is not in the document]
MarkMed: Ready! Fix that in commit 3068cfc8c6082ae066003baabf10cc969bd375b7

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,445 @@
...
+   * The Currecnt value indicanding in the gauge
+   *
+   */
+  @Prop() currentValue: number = this.minValue;
+
+  /**
+   * This allows specifying the width of the circumference _(When gauge is Circle type)_ and the width of the bar _(When gauge is Line type)_ in % relative the component size.
+   *
+   */
+  @Prop() thickness = 10;
```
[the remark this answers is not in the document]
MarkMed: This property cannot be a CSS variable because it used within the code for calculations purposes.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+@Component({
+  shadow: false,
+  styleUrl: "gauge.scss",
+  tag: "gx-gauge"
+})
+export class Gauge implements IComponent {
+  @Element() element: HTMLElement;
+
+  @Event() gxGaugeDidLoad: EventEmitter;
```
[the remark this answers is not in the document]
MarkMed: Comment added in commit af4ba8a42626d1f525a9cb38ede6412ba416ddb2

src/components/gauge/readme.md
MarkMed: Sample added in commit 3068cfc8c6082ae066003baabf10cc969bd375b7

src/components/map/map.tsx
MarkMed: I make the changes and include the file by mistake... 😓 
I revert the changes in the commit b8bcfedbb990f986a4b074127c3ee8433026d7d2

src/components/gauge/gauge.e2e.ts
MarkMed: Test added in commit e372c6e70aaaae9d521022815fbda83e2af2fe45

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,199 @@
+@import "../common/_base";
+
+gx-gauge {
```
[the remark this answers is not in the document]
MarkMed: In various commit I implement the SASS variables. e0d7cd94a0aaef01294495c42d49f864c9906d78 , 3d85f4459ec909f6c2c8f5b459cc23fa15b86236 and 9372632eddeb7dc54d70767fc48150fe644d0c18 (i don't know if there is another commit related with this)

src/components/gauge/gauge.tsx
MarkMed: This I fix in the 4700ea25a372abf35b3dfc54b371e0188315facd commit

src/components/gauge/gauge.tsx
MarkMed: It needs that function to going adding values to each range. It is of critical usage within the drawing process!

src/components/gauge/gauge.tsx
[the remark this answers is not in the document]
ncardeli: Still, I feel that, as it's used to calculate the value of other CSS properties, this is a good candidate for CSS variable, using the `calc()` function.

src/components/gauge/gauge.tsx
ncardeli: Yes, I understand, I'm just not sure if having a function for such a simple operation was needed.
Also remember that the function is evaluated every time the method is called, and, where it's called, there's a longer operation, that migth make more sense to encapsulate, to simplify reading and understanding.

src/components/gauge/gauge.tsx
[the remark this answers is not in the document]
MarkMed: 😓 
Look for example the next line:
```typescript
(this.minimumSize <= 300 && this.thickness <= 25) || !this.styleShadow
   ? "none"
   : ""
```
It needs the thickness just right there, next to the `minimumSize` to evaluate if it active the shadows or not.
For example, if the user set shadows but the control is very little. There, it will evaluate if set shadows or not. It will depend on the

src/components/gauge/gauge.tsx
MarkMed: Magic Numbers fixed in commits af4ba8a42626d1f525a9cb38ede6412ba416ddb2 beb727049f504fe25517716b0b1e6f789ef7442b ef4c6c284ffd4503c0216ca81a8c927ca297e085

src/components/gauge/gauge.tsx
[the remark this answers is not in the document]
MarkMed: Yes! That feature was added in commits 0e59ef3fd6ce41b037f6bbcb8de2125a866df0ab 3d85f4459ec909f6c2c8f5b459cc23fa15b86236 9372632eddeb7dc54d70767fc48150fe644d0c18

MarkMed:
PR Canceled.
A new PR will be created from a branch from this repository

PR #22, opened 2019-05-23
Add/rating component

MarkMed:


src/components/rating/rating.tsx
```diff
@@ -0,0 +1,177 @@
+// https://www.cssscript.com/demo/radio-button-based-star-rating-system/
```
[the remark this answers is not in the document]
MarkMed: That was an error. In the beginning, I drop those comments to have a direct link to go to the places where I take a base.

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,177 @@
...
+  @Prop() disabled = false;
+
+  handleChange: (UIEvent: any) => void;
+
+  @Element() element: HTMLElement;
+
+  /**
+   * The control id. Must be unique per control!
+   */
+  @Prop() id: string;
```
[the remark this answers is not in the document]
MarkMed: 😮 
I will fix that 😉

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private addStarsRating() {
+    const stars = [];
+    for (let i = 0; i < 5; i++) {
+      stars.push(
+        <svg
+          class="rating"
+          viewBox="0 0 100 100"
+          star-number={1 + i}
```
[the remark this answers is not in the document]
MarkMed: `star-number` in the `svg` element is not a standard attribute. Would you prefer to use the `data-` prefix to use custom attributes?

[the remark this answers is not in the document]
ncardeli: It could be an alternative. I just wonder if we really need an attribute in the element. Are there other alternatives to know which star is related to the svg element? Maybe the index in the parent child elements?

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,177 @@
...
+            min="0"
+            max={this.maxValue}
+            step="1"
+            value={this.value}
+          />
+          <div class="svgContainer score">{this.addStarsScore()}</div>
+        </div>
+      ) : (
+        // tslint:disable-next-line:no-console
+        console.error(
```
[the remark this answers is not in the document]
MarkMed: Do you want that it reports the exact error? Because I put that error log for show that there is something wrong with the value of `max-value` attribute or `value` attribute

src/components/rating/rating.e2e.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(element).toBeTruthy();
+    expect(await element.find(".svgContainer.rating")).toBeTruthy();
+  });
+
+  it("should be in score mode when readonly attr is true", async () => {
+    element.setAttribute("readonly", "true");
+    element.setAttribute("max-value", "5");
+    element.setAttribute("value", "5");
+    await page.waitForChanges();
+    expect(await element.find(".svgContainer.score")).toBeTruthy();
```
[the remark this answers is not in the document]
MarkMed: That test is for check id the component is in 'score mode'. Just check if an inner element is with a specific CSS class. The `max-value` and `value` attribute must are defined to help run the control.
The next test checks how impacts variations in the values of those attributes.

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,177 @@
...
+      return this.maxValue && this.value ? (
+        <div>
+          <input
+            type="range"
+            min="0"
+            max={this.maxValue}
+            step="1"
+            value={this.value}
+          />
+          <div class="svgContainer score">{this.addStarsScore()}</div>
```
ncardeli: `score` class is used both for a `svg` element and for this `div`. It might be a good idea to have different class names.

src/components/rating/rating.tsx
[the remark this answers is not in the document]
ncardeli: Yes, if possible, please.

ncardeli:
There's a lint error, can you take a look at it please?

src/components/rating/readme.md
```diff
@@ -0,0 +1,50 @@
+# gx-rating
+
+The `gx-rating` allows displaying scores or rate items using stars.
+
+E.g.:
+
+<div>
```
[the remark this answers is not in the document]
MarkMed: Oh! I thought that you wanted a live example of how the component will see. You want a code example of how the user must use the control Right?
Something like the following example, right?

```
<code>
   <code-son/>
</code>
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <div>
+        <input
+          type="range"
+          min="0"
+          max={this.readonly ? this.maxValue : 5}
+          step="1"
+          value={this.readonly ? this.value : this.ratingScore}
+        />
+        <div
+          class={`svgContainer ${this.readonly ? "score" : "rating"}`}
```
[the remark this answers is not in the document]
MarkMed: 😮 I did not know this way to set the attrs!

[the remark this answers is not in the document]
MarkMed: Yeeeap. You are right. It's redundant to set a class. There's just one `<div>` within. I will remove it.

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * The 'input' event is emitted when a change to the element's value is committed by the user.
+   */
+  @Event() input: EventEmitter;
+
+  onClick(event: UIEvent) {
+    const element = event.target as HTMLElement;
+    const score =
+      element.nodeName === "polygon"
+        ? Array.from(element.parentElement.parentElement.children).indexOf(
```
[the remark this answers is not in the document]
MarkMed: I thought that too.
The problem is that the parent of the target element is undefined. Clicking in a star the `onClick` event is triggered but the element target can be the `<polygon>` element or the `<svg>` element. The variable where the parent element will be saved will not have a specific name like: 
 ```typescript
const svgElement = element.parentElement;;
```
Doing what you correctly recommend it will be something like:
```typescript
onClick(event: UIEvent) {
    const element = event.target as HTMLElement;
    const targetParent = element.parentElement;
    element.nodeName === "polygon"
        ? Array.from(targetParent.parentElement.children).indexOf(
            targetParent
          ) + 1
        : Array.from(targetParent.children).indexOf(element) + 1;
    this.ratingScore = score;
    this.input.emit(this);
}
```

src/components/rating/readme.md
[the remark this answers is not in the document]
ncardeli: Yes, an example of how the control is used. Thanks!

src/components/rating/readme.md
```diff
@@ -0,0 +1,50 @@
...
+    <svg viewBox="0 0 100 100" width="40" heigth="40" style=" filter: drop-shadow(0 0 2px #e4ac13);"><polygon fill="#e4ac13" points="50,0 15,95 100,35 0,35 85,95"/></svg>
+    <svg viewBox="0 0 100 100" width="40" heigth="40" style=" filter: drop-shadow(0 0 2px #e4ac13);"><polygon fill="#e4ac13" points="50,0 15,95 100,35 0,35 85,95"/></svg>
+    <svg viewBox="0 0 100 100" width="40" heigth="40" style=" filter: drop-shadow(0 0 2px #e4ac13);"><polygon fill="#e4ac13" points="50,0 15,95 100,35 0,35 85,95"/></svg>
+    <svg viewBox="0 0 100 100" width="40" heigth="40" style=" filter: drop-shadow(0 0 2px #e4ac13);"><polygon fill="#e4ac13" points="50,0 15,95 100,35 0,35 85,95"/></svg>
+    <svg viewBox="0 0 100 100" width="40" heigth="40" style=" filter: drop-shadow(0 0 2px #e4ac13);"><polygon fill="#e4ac13" points="50,0 15,95 100,35 0,35 85,95"/></svg>
+</div>
+
+Use `gx-rating` with `readonly = true` to make available the 'score mode' and set values for `max-value` and `value` attributes.
```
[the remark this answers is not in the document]
MarkMed: Aahhhh I do not how to call to it. In this component you have two "modes":
- The **_"Rating mode"_**.
	> Where the `readonly` property has `false` value and the user is available to vote and click on the stars.
- The **_"Score mode"_**.
	> Where the `readonly` property has `true` value and the user will only be able to see a score.

And I do not know how to explain that in a simple name.
Perhaps include this explanation in the `readme.md` file, what do you think?

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,154 @@
...
+    this.ratingScore = score;
+    this.input.emit(this);
+  }
+
+  /**
+   * Returns the id of the inner `input` element (if set).
+   */
+  @Method()
+  async getNativeInputId() {
+    return this.render();
```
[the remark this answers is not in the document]
MarkMed: should return a DOM element or the Id of that DOM element?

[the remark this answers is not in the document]
ncardeli: Sorry, I meant the id of the DOM element. Take a look at other components that implement that method.

src/components/rating/rating.scss
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+  }
+  input {
+    display: none;
+  }
+  div.rating {
+    @include widthHeightsize(max-content, max-content);
+    @include flex(row, space-between, center, wrap);
+    &[data-score="1"] {
```
ncardeli: All the `[data-score="x"]` blocks are almost the same. I feel we could do better here to avoid code repetition.

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { IFormComponent } from "../common/interfaces";
+@Component({
+  shadow: false,
+  styleUrl: "rating.scss",
+  tag: "gx-rating"
+})
+export class Rating implements IFormComponent {
+  inputId: string;
+
+  starShape = <polygon points="50,0 15,95 100,35 0,35 85,95" />;
```
ncardeli:
```suggestion
  private starShape = <polygon points="50,0 15,95 100,35 0,35 85,95" />;
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  svgViewport = {
```
ncardeli:
```suggestion
  private svgViewport = {
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+  Prop
+} from "@stencil/core";
+import { IFormComponent } from "../common/interfaces";
+@Component({
+  shadow: false,
+  styleUrl: "rating.scss",
+  tag: "gx-rating"
+})
+export class Rating implements IFormComponent {
+  inputId: string;
```
ncardeli:
```suggestion
  private inputId: string;
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  starShape = <polygon points="50,0 15,95 100,35 0,35 85,95" />;
+
+  svgViewport = {
+    viewBox: "0 0 100 100"
```
ncardeli:
```suggestion
    private viewBox: "0 0 100 100"
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+          {...this.svgViewport}
+          onClick={ev => this.onClick(ev)}
+        >
+          {this.starShape}
+        </svg>
+      );
+    }
+    return stars;
+  }
+  private renderStarsScore() {
```
ncardeli:
```suggestion

  private renderStarsScore() {
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+              score: this.readonly
+            }}
+            data-score={this.value !== 0 ? this.value : undefined}
+          >
+            {this.readonly ? this.renderStarsScore() : this.renderStarsRating()}
+          </div>
+        </div>
+      );
+    } else {
+      if (!!this.maxValue) {
```
ncardeli: There's no need to double negate when checking for falsy values.

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+        // tslint:disable-next-line:no-console
+        console.error(
+          "Incongruence between values. 'value' is higher than 'max-value'.",
+          this
+        );
+      } else {
+        // tslint:disable-next-line:no-console
+        console.error("'max-value' has not a value setted ", this);
```
ncardeli:
```suggestion
        console.error("'max-value' has not a value set ", this);
```

src/components/rating/rating.tsx
```diff
@@ -0,0 +1,183 @@
...
+          >
+            {this.readonly ? this.renderStarsScore() : this.renderStarsRating()}
+          </div>
+        </div>
+      );
+    } else {
+      if (!!this.maxValue) {
+        // tslint:disable-next-line:no-console
+        console.error(
+          "Incongruence between values. 'value' is higher than 'max-value'.",
```
ncardeli:
```suggestion
          "'value' cannot be higher than 'max-value'.",
```

src/components/rating/rating.tsx
ncardeli: Reflecting on this log messages, I'm wondering if we shouldn't be adding the element to the log line, so the user has context as to where the error was originated.

src/components/rating/readme.md
```diff
@@ -0,0 +1,68 @@
+# gx-rating
+
+The `gx-rating` allows displaying scores or rate items using stars.
```
ncardeli:
```suggestion
The `gx-rating` component allows displaying scores or rate items using stars.
```

src/components/rating/readme.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `gx-rating` with `readonly` disabled or false:
```
ncardeli:
```suggestion
## Samples

- `gx-rating` with `readonly` disabled or false:
```

src/components/rating/rating.tsx
MarkMed: It allows setting the variables as private, not the properties of objects.

src/components/rating/rating.tsx
MarkMed: fixed!

src/components/rating/rating.tsx
MarkMed: fixed!

src/components/rating/rating.tsx
MarkMed: I feel so dumb... 😆 😭 
Fixed!

src/components/rating/rating.tsx
MarkMed: fixed!

src/components/rating/rating.tsx
MarkMed: fixed!

src/components/rating/rating.tsx
MarkMed: Ready!
It will print the gx element from the DOM

src/components/rating/readme.md
MarkMed: 👌

src/components/rating/readme.md
MarkMed: That's true... I forgot include a title like _Usage Example_ or something like that. 😅

MarkMed: Ready!

ncardeli:
This looks great! 💯

PR #25, opened 2019-06-05
Lazy loading for gx-image component

ncardeli:
Added support for lazy loading images in the `gx-image` component.
It's enabled by default, but can be disabled using the `lazy-load` attribute.
Images below the fold aren't loaded until they are close to or in the viewport.
`lazysizes` was used to implement this feature as it's a fast, widely used, lightweight library for lazy loading images, that supports all major browsers and IE9 and up.

When we add support for multiple densities in the future, using the `srcset` attribute, this library will be useful too, as it also support lazy loading the images specified in the `srcset` attribute.

A CSS animation with a spinning circle is shown while the image is being loaded.

Also added support for specifying a class for highlighting, in theming mixin.


ggallotti:
Looks good!
Does this control supports "Image Loading Indicator"?

ncardeli:
> Looks good!
> Does this control supports "Image Loading Indicator"?

Yes, it does! There's a CSS animation called gx-image-loading used for animating a <span> element that's only visible while loading.

PR #26, opened 2019-06-05
Fix race condition when initializing lazySizes

ncardeli:
Changed the moment when lazySizes config is initialized. Now it's done immediately after the component's class is defined, instead of the constructor.
This is to avoid a race condition when `gx-image` components are created dynamically.

### DIFF
--- a/src/components/image/image.tsx
+++ b/src/components/image/image.tsx
@@ -19,12 +19,6 @@ export class Image
     IDisableableComponent,
     IVisibilityComponent,
     IClickableComponent {
-  constructor() {
-    lazySizes.cfg.lazyClass = LAZY_LOAD_CLASS;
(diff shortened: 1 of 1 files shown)


src/components/image/image.tsx
```diff
@@ -111,8 +107,17 @@ export class Image
     ];
     return body;
   }
+
+  private shouldLazyLoad(): boolean {
+    const img: HTMLImageElement = this.element.querySelector("img");
+    return this.lazyLoad && (!img || img.classList.contains(LAZY_LOAD_CLASS));
```
ggallotti: Minor suggestion: We could return immediately if lazyLoad is false. So we do not querySelector unnecessary.

ncardeli: Great catch! Thanks!

PR #40, opened 2019-07-30
Add event/map component

MarkMed:
`gx-map` with the `mapClick` event that returns the coordinate where the click/tap has been done in the map layout


src/components/map/map.scss
```diff
@@ -1,10 +1,16 @@
...
+@mixin widthHeightsize($valueX, $valueY) {
+  height: $valueY;
+  width: $valueX;
+}
 gx-map {
   @include visibility(block);
-  .gxMap {
-    width: 100%;
-    height: 360px;
+  .gxMapContainer {
```
ncardeli: This is ok, but next time, consider using `display: flex` on the parent element and `flex: 1` in the child element to accomplish the same behavior (instead of 100% for both width and height).

The main advantage is that using that technique, you can set margins to the child element without overflowing the parent element.

PR #59, opened 2019-08-23
Do not bind 'Click' event handlers when disabled

ggallotti:
Only set OnClick handlers when needed (not disabled)


src/components/renders/bootstrap/navbar-link/navbar-link-render.tsx
```diff
@@ -27,7 +23,7 @@ export class NavBarLinkRender implements IRenderer {
           [this.component.cssClass]: !!this.component.cssClass
         }}
         href={this.component.href}
-        onClick={this.handleClick.bind(this)}
+        onClick={!this.component.disabled ? this.handleClick.bind(this) : null}
```
ncardeli: Just nitpicking..
```suggestion
        onClick={!this.component.disabled && this.handleClick.bind(this)}
```

PR #75, opened 2019-09-11
Use a custom build of Bootstrap CSS

ncardeli:
Instead of using the standard distribution of Bootstrap, we are creating our own custom build, using only those classes used by the library controls.
To achieve this, we are creating the custom Bootstrap build importing the `scss` files from Bootstrap's source.

This adds a small overhead in the build process, by increasing the amount of work the SASS build step has to perform, but it pays off by letting us save some KBs in the final bundle.

Additionally, it means that when creating new controls, we should verify that the required Bootstrap `scss` files for implementing the control, are imported.


ncardeli:
> Why don't we include sass single dependencies in each web component?
> Instead of the full bootstrap?

Shared dependencies would be included in each component, resulting in bigger bundles for each component.

ncardeli:
I also removed the Font Awesome dependency ✂️ 
This one will remove some KBs of CSS, and will also remove some roundtrips to get the font files. 🚀

PR #103, opened 2020-03-20
Fix to image control: removed setting that hides when there is no src

ag-silva:
Removed setting that hides the control when there is no src, because it conflicts with hidden setting in Angular when it loads for the first time.


MarkMed:
What happens now if there is no src specified?
I think that it would be great to load a default img or a generic img in those cases.

E.g.:
```html
<gx-img auto-grow="false" width="200" height="200"></gx-img>
```
---
image

PR #104, opened 2020-03-25
Add/popup marker

MarkMed:
This PR is to add the new Popup functionality for map markers

This is a summary of all commits from [MarkerPopup](https://github.com/genexuslabs/web-controls-library/tree/add/MarkerPopup) branch

### DIFF
--- a/src/components/map-marker/map-marker.tsx
+++ b/src/components/map-marker/map-marker.tsx
@@ -10,6 +10,8 @@ import { Component as GxComponent } from "../common/interfaces";
 import { divIcon, marker } from "leaflet/dist/leaflet-src.esm";
 import { parseCoords } from "../common/coordsValidate";
 
+const MAX_POPUP_SIZE_FACTOR = 0.83;
+
 @Component({
   shadow: false,
   styleUrl: "map-marker.scss",
@@ -94,11 +96,11 @@ export class MapMarker implements GxComponent {
       "[class='popup-data-container']"
     );
     if (popupContainerEl.firstElementChild !== null) {
-      const marginProportion = 83 / 100;
       const maxPopupSize = {
         height:
-          document.querySelector(".gxMap").clientHeight * marginProportion,
-        width: document.querySelector(".gxMap").clientWidth * marginProportion
+          document.querySelector(".gxMap").clientHeight * MAX_POPUP_SIZE_FACTOR,
+        width:
+          document.querySelector(".gxMap").clientWidth * MAX_POPUP_SIZE_FACTOR
       };
       this.markerInstance.bindPopup(popupContainerEl, {
         keepInView: true,
@@ -130,10 +132,6 @@ export class MapMarker implements GxComponent {
   }
 
   componentDidUpdate() {
-    const halfIconSizes = {
-      height: this.iconHeight / 2,
-      width: this.iconWidth / 2
-    };
     const coords = parseCoords(this.coords);
     if (coords !== null) {
       this.markerInstance.setLatLng(coords);
@@ -150,7 +148,7 @@ export class MapMarker implements GxComponent {
         iconAnchor: [this.getHalfSizes().width, this.iconHeight],
         popupAnchor: [0, -this.getHalfSizes().height],
         iconSize: [this.iconWidth, this.iconHeight],
-        tooltipAnchor: [0, -halfIconSizes.height]
+        tooltipAnchor: [0, -this.getHalfSizes().height]
       })
     );
     this.setPopup();
--- a/src/components/map-marker/map-marker.tsx
+++ b/src/components/map-marker/map-marker.tsx
@@ -70,15 +70,17 @@ export class MapMarker implements GxComponent {
   @Event() gxMapMarkerDeleted: EventEmitter;
 
   private setPopup() {
-    const popupHtml = this.element.querySelector("[class='popupHtml']");
-    if (popupHtml.innerHTML) {
+    const popupContainerEl = this.element.querySelector(
+      "[class='popup-data-container']"
+    );
+    if (popupContainerEl.innerHTML) {
       const marginProportion = 83 / 100;
       const maxPopupSize = {
         height:
           document.querySelector(".gxMap").clientHeight * marginProportion,
         width: document.querySelector(".gxMap").clientWidth * marginProportion
       };
-      this.markerInstance.bindPopup(popupHtml, {
+      this.markerInstance.bindPopup(popupContainerEl, {
         keepInView: true,
         maxHeight: maxPopupSize.height,
         maxWidth: maxPopupSize.width,
@@ -160,7 +162,7 @@ export class MapMarker implements GxComponent {
 
   render() {
     return (
-      <div class="popupHtml">
+      <div class="popup-data-container">
         <slot />
       </div>
     );
--- a/src/components/map-marker/map-marker.tsx
+++ b/src/components/map-marker/map-marker.tsx
@@ -142,7 +142,6 @@ export class MapMarker implements GxComponent {
       );
       this.markerInstance.setLatLng([0, 0]);
     }
-    // this.setIconSrc();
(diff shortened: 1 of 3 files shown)


ncardeli:
Please provide a description of what the PR.

Thanks!

MarkMed:
Ok.
This PR is to add the new Popup functionality of marker

src/components/map-marker/map-marker.tsx
```diff
@@ -62,16 +69,38 @@ export class MapMarker implements GxComponent {
    */
   @Event() gxMapMarkerDeleted: EventEmitter;
 
+  private setPopup() {
+    const popupHtml = this.element.querySelector("[class='popupHtml']");
```
ncardeli: Nothing big here, just wanted to mention that `popupHtml` doesn't really describe the content of the variable, right? I mean, it's a reference to a DOM element, not HTML.
It confused me at first a little.

src/components/map-marker/map-marker.tsx
```diff
@@ -62,16 +69,38 @@ export class MapMarker implements GxComponent {
...
+    if (popupHtml.innerHTML) {
```
ncardeli: What are we checking with this condition?

src/components/map-marker/map-marker.tsx
```diff
@@ -62,16 +69,38 @@ export class MapMarker implements GxComponent {
...
+      const marginProportion = 83 / 100;
```
ncardeli: Magic numbers 👀

src/components/map-marker/map-marker.tsx
```diff
@@ -62,16 +69,38 @@ export class MapMarker implements GxComponent {
...
+  }
+
   componentDidLoad() {
-    const halfIconWidth = this.iconWidth / 2;
+    const halfIconSizes = {
+      height: this.iconHeight / 2,
+      width: this.iconWidth / 2
+    };
     const coords = parseCoords(this.coords);
     if (coords !== null) {
```
ncardeli: Shouldn't  we emit the `gxMapMarkerDidLoad` inside this `if` block?

src/components/map-marker/map-marker.tsx
```diff
@@ -81,20 +110,28 @@ export class MapMarker implements GxComponent {
...
     }
     this.gxMapMarkerDidLoad.emit(this.markerInstance);
   }
 
   componentDidUpdate() {
-    const halfIconWidth = this.iconWidth / 2;
+    const halfIconSizes = {
+      height: this.iconHeight / 2,
+      width: this.iconWidth / 2
+    };
```
ncardeli: This is repeated in the `componentDidLoad` event. Can we encapsulate this?

src/components/map-marker/map-marker.tsx
```diff
@@ -62,16 +69,38 @@ export class MapMarker implements GxComponent {
...
       this.markerInstance = marker(coords, {
         icon: divIcon({
           className: this.markerClass,
-          iconAnchor: [halfIconWidth, this.iconHeight],
+          iconAnchor: [halfIconSizes.width, this.iconHeight],
+          popupAnchor: [0, -halfIconSizes.height],
           iconSize: [this.iconWidth, this.iconHeight],
-          tooltipAnchor: [0, -28]
+          tooltipAnchor: [0, -halfIconSizes.height]
         })
```
ncardeli: I feel there's a refactoring opportunity here. This code block is repeated several times.

src/components/map-marker/readme.md
```diff
@@ -3,6 +3,30 @@
 This element represents a marker inside a `<gx-map>` element.
 `coords` attribute must always have a value. If value is not specified, the marker will be initialized with default values and it will be drawn at `0, 0` coordinates.
 
+## Using Sample
```
ncardeli:
```suggestion
## Usage Sample
```

src/components/map-marker/map-marker.tsx
MarkMed: ```popupHtml``` is about the HTML that will be inside in popup of marker. I need a way to check if there is any content inside ```<gx-map-marker>```, so, I use the container ```<div class="popupHtml">``` to check out the content. Perhaps I should rename it to _"popupContent"_ or _"popupHTMLContainer"_ ? Wht do u suggest?

ncardeli: If `popupHtml` references an element, a better name could be `popupEl` or `popupContainerEl` if you want to be more specific about what the element is for.

MarkMed: Mmm... I must do reference (using a variable) to the popup data container ```<div>```. To that ```<div>```, I will rename to its class as _"popupDataContainer"_, and the variable that refers to that element I will call it as _"popupContainerEl"_ . wht do you think about it?

ncardeli: +1 with the variable name 👍 
As for the class name, I welcome the renaming. However, we are using dash separated, lower case names for classes, like `popup-data-container`, so please reconsider the new name.

src/components/map-marker/map-marker.tsx
MarkMed: basically, we check if users/developers have introduced some content inside ```<gx-map-maker>``` . If there is  any content (html), It calls leaflet method to add that content into the popup of marker (marker created by leaflet)

ncardeli: Please consider that accessing the `innerHtml` property comes at a cost. If your just checking if it has content, there are other cheaper alternatives, like `firstElementChild !== null`.

src/components/map-marker/map-marker.tsx
MarkMed: No no 😅
```marginProportion``` is a variable to avoid that magic number ```83/100```
I decided to leave it like that because is more mathematically intuitive.
Don't need to be a super mathy person to realize that ```83/100``` is 83% of something, and I think that ```83/100``` is better than type just ```0.83```...  it's more mathematically intuitive...

Wht do u think that i shall do?

src/components/map-marker/map-marker.tsx
MarkMed: Yep! I just realize that piece of code is repeated twice in the ```componentDidLoad``` method.
I'll fix that!

src/components/map-marker/map-marker.tsx
MarkMed: fixed on f956dec commit

src/components/map-marker/map-marker.tsx
MarkMed: ```gxMapMarkerDidLoad``` must be emitted after process all data and marker structure. ```gxMapMarkerDidLoad``` is successfully emitted at the end of ```componentDidLoad``` method.

src/components/map-marker/readme.md
MarkMed: fixed on f956dec commit

src/components/map-marker/map-marker.tsx
MarkMed: done in 7dfb8c6 👌

src/components/map-marker/map-marker.tsx
MarkMed: done in 7dfb8c6

src/components/map-marker/map-marker.tsx
MarkMed: done in 7dfb8c6

src/components/map-marker/map-marker.tsx
ncardeli: .83 is fine.
You could also leave this constant value outside the class, for example:

const MAX_POPUP_SIZE_FACTOR = .83;

src/components/map-marker/map-marker.tsx
ncardeli: Sorry, my bad.

PR #112, opened 2020-05-26
Add/selection layer

MarkMed:
Early PR to working on corrections and of extra details request of ```Selection Layer``` feature


src/components/map/map.tsx
```diff
@@ -106,6 +119,18 @@ export class Map implements GxComponent {
    */
   @Event() mapClick: EventEmitter;
 
+  /**
+   * Emmits when the map is being moving while selection layer is active.
```
ncardeli, line 123:
```suggestion
   * Emmited when the map is being moved, if selection layer is active.
```

src/components/map/map.tsx
```diff
@@ -106,6 +119,18 @@ export class Map implements GxComponent {
...
+   *
+   */
+  @Event() selectionInput: EventEmitter;
+
+  /**
+   * Emmits when the map stoped from being moving while selection layer is active.
```
ncardeli, line 129:
```suggestion
   * Emmited when the map stops being moved, if selection layer is active.
```

src/components/map/map.tsx
```diff
@@ -106,6 +119,18 @@ export class Map implements GxComponent {
...
+   *
+   */
+  @Event() selectionChange: EventEmitter;
+
   /**
    * Emmits when user location coords have been changed.
```
ncardeli, line 135:
```suggestion
   * Emmited when the user location coords change.
```

src/components/map/map.tsx
```diff
@@ -153,6 +198,15 @@ export class Map implements GxComponent {
...
 
+  private isSelectionMarkerSlot(): { exist: boolean; elem: Element } {
+    const slot = this.element.querySelector("[slot='selection-layer-marker']");
+    if (slot !== null) {
+      return { exist: true, elem: slot };
+    } else {
+      return { exist: false, elem: null };
+    }
+  }
+
```
ncardeli, line 209: Could we simplify this function?
I feel that if the function just returns the slot, the caller can decide if it exists or not by checking if the function returns an object or null. In that case, I'd change the name to 
`getSelectionMarkerSlot`.

src/components/map/map.tsx
```diff
@@ -169,6 +223,50 @@ export class Map implements GxComponent {
     }
   }
 
+  private updateSelectionMarkerPosition() {
+    const centerCoords = this.map.getCenter();
+    this.centerCoords = `${centerCoords.lat},${centerCoords.lng}`;
+    this.selectionMarker.setAttribute("coords", this.centerCoords);
```
ncardeli, line 229: We could declare `selectionMarker` with a more specificc type, taking advantage of `HTMLGxMapMarkerElement` (is a type automatically generated by Stencil in `components.d.ts`).
That way, we could set the `coords` property, instead of `setAttribute`:

```
this.selectionMarker.coords = this.centerCoords;
````

File: src/components/map/map.tsx [hunk 6]
```diff
@@ -169,6 +223,50 @@ export class Map implements GxComponent {
...
+  }
+
+  private selectionLayerEvents() {
```
ncardeli, line 232: Maybe we can give this method a better name, that reflects the action taken with the selection layer events.

src/index.html
```diff
@@ -9,8 +9,106 @@
     <title>GeneXus Web Controls Library test page</title>
     <script type="module" src="/build/gx-web-controls.esm.js"></script>
     <script nomodule src="/build/gx-web-controls.js"></script>
+    <style>
```
ncardeli, line 12: Let's leave index.html empty. It's great if you want to create a sample page, but it would be better if you create a new html page inside the `samples` directory, under the root directory.

MarkMed: yes! I forgot to do that. thx for remind it me! :D

src/components/map/map.tsx
MarkMed: Fixed in f3a1266

src/components/map/map.tsx
MarkMed: Fixed in f3a1266

src/components/map/map.tsx
MarkMed: Fixed in f3a1266

src/index.html
MarkMed: Ready! Fixed in f3a1266

src/components/map/map.tsx
MarkMed: Oh! So convenient!
I had a lot of obstacles with that problem... `selectionMarker: Element` was the final conclusion
I will try your solution! Thx!

MarkMed: Mmm... There is a conflict here.

If I make `selectionMarker: HTMLGxMapMarkerElement` there will be a type error when I search for slot with `this.element.querySelector("[slot='selection-layer-marker']")`
The querySelector returns an `Element` type 😣

image

src/components/map/map.tsx
MarkMed: Yas. it needs more clues about its purpose, but... what name? `selectionLayerMapEventsSubscriber`?

Let's see: That method evaluates the state of selectionLayer (if it is `true` or `false`) and subscribes or unsubscribes the map to events emitted by leaflet and give certain behavior depending in the event listened
So, what do you think to call it `selectLay_EventManager`?

src/components/map/map.tsx
MarkMed: I can't believe that... redundant code 🤦‍♂️ 
I will immediately fix that disrespect.

MarkMed: Fixed! See changes in fd50e4d

src/components/map/map.tsx
ncardeli: Try this: 

`const slot = this.element.querySelector<HTMLGxMapMarkerElement>("[slot='selection-layer-marker']")`

src/components/map/map.tsx
ncardeli: I usually start methods with a verb. In this case, I'd use: `registerSelectionLayerEvents`

src/components/map/map.tsx
MarkMed: 😮 
i didn't know that. 
I saw it in  the popup dialog, but I did not want to try...
I will try! 😄

src/components/map/map.tsx
MarkMed: defining name of methods with a verb 🤔 That was a good one. I will remember that.

MarkMed: renamed!
Look in comit be8cd25

src/components/map/map.tsx
MarkMed: It works! The latest tests have passed successfully with these new changes.
See new changes in ad11b7e

MarkMed:
@ncardeli we cannot merge yet!
I discover an error 😞 

When `this.selectionLayer = false` it suppose to render the map normally. But by doing test I discover that automatically inserts the generic `<gx-map-marker marker-class="gx-default-selection-layer-icon"/>`
I don't know why it does that...
@ncardeli please check if the `render()` section is everything right, perhaps there is an error that I didn't know _(the inline conditional and ternary conditional)_

I will keep searching for the bug. 🧐

src/components/map/map.tsx
```diff
@@ -268,6 +380,16 @@ export class Map implements GxComponent {
...
+        {this.selectionLayer && this.isSelectionLayerSlot ? (
+          <slot name="selection-layer-marker" />
+        ) : (
+          <gx-map-marker
+            marker-class="gx-default-selection-layer-icon"
+            icon-width="30"
+            icon-height="30"
+            coords={this.centerCoords}
+          ></gx-map-marker>
+        )}
```
ncardeli, line 392: This condition is what makes the render method introduce the default maker when `selectionLayer` is false

```suggestion
        this.selectionLayer && {this.isSelectionLayerSlot ? (
          <slot name="selection-layer-marker" />
        ) : (
          <gx-map-marker
            marker-class="gx-default-selection-layer-icon"
            icon-width="30"
            icon-height="30"
            coords={this.centerCoords}
          ></gx-map-marker>
        )}
```

MarkMed: Interesting suggestion 🤔 
I will apply those changes too! But I found the bug in the handler of  `MapMarkerDidLoad` event. I will fix that and apply your suggestion!

MarkMed: It does not works. When you move the `{` the code behind.

ncardeli: Sorry, proposed the change but didn't properly verify it.

Let's try again:

```
        {this.selectionLayer && (this.isSelectionLayerSlot ? (
          <slot name="selection-layer-marker" />
        ) : (
          <gx-map-marker
            marker-class="gx-default-selection-layer-icon"
            icon-width="30"
            icon-height="30"
            coords={this.centerCoords}
          ></gx-map-marker>
        ))}
```

MarkMed: This works! I made another solution, but I like yours.
My solution was:
```typescript
        let selectionMarkerElement;
	if(this.selectionLayer){
		selectionMarkerElement = this.isSelectionLayerSlot ? (
			<slot name="selection-layer-marker" />
		  ) : (
			<mkm-map-marker
			  marker-class="mkm-default-selection-layer-icon"
			  icon-width="30"
			  icon-height="30"
			  coords={this.centerCoords}
			></mkm-map-marker>
		  )
	}
```
and in the return:
```typescript
        {this.selectionLayer && selectionMarkerElement}
```

MarkMed: fixed in 5262096

src/components/map/map.tsx
```diff
@@ -29,7 +29,8 @@ const RECOMMENDED_MAX_ZOOM = 20;
   tag: "gx-map"
 })
 export class Map implements GxComponent {
-  private watchPositionId: number;
+  private centerCoords: string;
```
ncardeli, line 32: Why do we have `centerCoords` and `center` properties. It seems both of them serve the same purpose. Could we refactor the code to only use `center`?
You can address this in another PR, if you prefer.

src/components/map/map.tsx
```diff
@@ -215,6 +316,9 @@ export class Map implements GxComponent {
         }
       );
     }
+    if (this.selectionLayer && this.getSelectionMarkerSlot().exist) {
+      this.isSelectionLayerSlot = true;
+    }
```
ncardeli, line 321:
```suggestion
    this.isSelectionLayerSlot = this.selectionLayer && this.getSelectionMarkerSlot().exist;
```

MarkMed:
Selection Layer made! Everything tested and ok!

PR #113, opened 2020-06-16
Vertical Grid improvements

ggallotti:
- Changes in Layout to support Virtual Scroll
- Improved Infinite Scrolling to support dynamic viewport
- Autogrow better support
- Better grid state handling (empty, loading, initial loading)


src/components/grid-base/grid-base.scss
```diff
@@ -5,23 +5,42 @@
...
+  & [slot="grid-content"] {
+    @include visibility(none);
+  }
+  & [slot="grid-empty-loading-placeholder"] {
+    flex: 1;
+    @include visibility(flex);
+    align-items: center;
+    justify-content: center;
+    text-align: center;
+    flex: 1;
```
ncardeli, line 44: Repeated `flex` 👁️

src/components/grid-infinite-scroll/grid-infinite-scroll.tsx
```diff
@@ -127,7 +134,13 @@ export class GridInfiniteScroll implements ComponentInterface {
   private isVisibleInViewport(el: HTMLElement): boolean {
     const rect = el.getBoundingClientRect();
     const elemTop = rect.top;
-    return elemTop >= 0 && elemTop <= window.innerHeight;
+    const element = this.getScrollListener();
+
+    return (
+      el.style.display !== "none" &&
```
ncardeli, line 140: Could we use the hidden attribute instead of the `display` property?

PR #115, opened 2020-06-22
Add/interactive image

MarkMed:
PR to add this new control to the library.
I think that this control should be added to the control library to be available and then gradually update with future features.


src/components/interactive-image/interactive-image.scss
```diff
@@ -0,0 +1,33 @@
...
+   * (auto by default)
+  */
+  --control-height: false;
+
+  /**
+	* @prop --control-width:
+	 * Set the width or the control. *Must be set before render the control*
+	 * (100% by default)
+	*/
+  --control-width: false;
```
ncardeli, line 21: Why do we need new custom CSS properties for setting the dimensions? Shouldn't we just use `width` and `height` like a regular HTML element?

src/components/interactive-image/interactive-image.scss
```diff
@@ -0,0 +1,33 @@
...
+
+  @include visibility(flex);
+  @include size(var(--control-width), var(--control-height));
+  background-position: center;
+  background-repeat: no-repeat;
+  overflow: hidden;
+
+  background-color: rgba(250, 0, 0, 0.2);
```
ncardeli, line 29: Why are we settings this default `background-color`?

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+  styleUrl: "interactive-image.scss",
+  tag: "gx-interactive-image"
+})
+export class InteractiveImage implements GxComponent {
+  @Element() element: HTMLGxInteractiveImageElement;
+
+  /**
+   * If this property is true, the user can copy the image to the clipboard (Only for iOS, see image below).
+   */
+  @Prop() enableCopyToClipboard: false;
```
ncardeli, line 15: We shouldn't add this property if we won't be doing anything with it in this PR.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Indicates how much you can enlarge an image. (Percentage) _Note: 100% = Normal size_.
+   */
+  @Prop() zoom = 100;
+
+  /**
+   * Indicates how much you can enlarge an image with reference to the original size of it or the size of the controller. (Percentage)
+   */
+  @Prop() maxZoomRelativeTo: number;
```
ncardeli, line 30: Let's remove this property from this PR. This is going to be our next feature, and we will work with it in our next PR.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * Lets you specify the "src" of the img.
```
ncardeli, line 33:
```suggestion
   * Lets you specify the image URL. Required.
```

src/components/interactive-image/interactive-image.scss
```diff
@@ -0,0 +1,33 @@
...
+	* @prop --control-width:
+	 * Set the width or the control. *Must be set before render the control*
+	 * (100% by default)
+	*/
+  --control-width: false;
+
+  @include visibility(flex);
+  @include size(var(--control-width), var(--control-height));
+  background-position: center;
+  background-repeat: no-repeat;
```
ncardeli, line 26: Do we need to set these background properties?

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  private correctZoomValue() {
+    if (this.zoom < 100) {
+      console.warn("Zoom value cannot be lower than 100");
+      this.zoom = 100;
+    }
+  }
+
+  private editingOverClass() {
```
ncardeli, line 104: Nitpicking here, but I feel we can come up with a better name for this method.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+    if (this.enableZoom && this.mouseOver) {
+      this.element.classList.add("zoom-over");
```
ncardeli, line 106: What does the `zoom-over` class do? I couldn't find it in the SASS file.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+      <img
+        style={
+          this.mouseOver
+            ? {
+                "object-position": this.mouseOver
+                  ? `${this.zoomedPositionX}px ${this.zoomedPositionY}px`
+                  : "0px 0px",
+                transform: this.mouseOver
+                  ? `scale(${this.zoom / 100})`
+                  : `scale(1)`
```
ncardeli, line 213: ❤️

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+    } else {
+      const img = this.element.querySelector("img");
+      this.removeEvent(img, zooming.over.withMouse, zooming.over.mouseBehaivor);
+      this.removeEvent(img, zooming.over.withTouch, zooming.over.touchBehaivor);
+      this.removeEvent(img, zooming.out.withMouse, zooming.out.behaivor);
+      this.removeEvent(img, zooming.out.withTouch, zooming.out.behaivor);
+    }
+  }
+
+  private correctZoomValue() {
```
ncardeli, line 97: How about `fixZoomValue`? I feel it sounds better.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+    callbackFunction
+  ) {
+    element.removeEventListener(eventToListen, callbackFunction);
+  }
+
+  private zoomFeature = {
+    over: {
+      withMouse: "mousemove",
+      withTouch: "touchmove",
+      mouseBehaivor: ev => {
```
ncardeli, line 124:
```suggestion
      mouseBehavior: ev => {
```

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+    out: {
+      withMouse: "mouseout",
+      withTouch: "touchend",
+      behaivor: () => {
+        this.mouseOver = false;
+        this.element.setAttribute("class", "hydrated");
+        this.element.style.backgroundPosition = `0 0`;
+      }
+    }
+  };
```
ncardeli, line 187: I'd like to understand the rationale behind the decision for creating this object. 
What are the advantages compared to having private class methods?

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+        this.zoomedPositionX = moveImgPostion.X;
+        this.zoomedPositionY = moveImgPostion.Y;
+      }
+    },
+    out: {
+      withMouse: "mouseout",
+      withTouch: "touchend",
+      behaivor: () => {
+        this.mouseOver = false;
+        this.element.setAttribute("class", "hydrated");
```
ncardeli, line 183: The `hydrated` shouldn't be manipulated by us. It's and internal class of Stencil.

src/components/interactive-image/interactive-image.tsx
```diff
@@ -0,0 +1,222 @@
...
+  render() {
+    this.correctZoomValue();
+    return (
+      <img
+        style={
+          this.mouseOver
+            ? {
+                "object-position": this.mouseOver
+                  ? `${this.zoomedPositionX}px ${this.zoomedPositionY}px`
+                  : "0px 0px",
```
ncardeli, line 210: No need for units when specifying `0`.
```suggestion
                  : "0 0",
```

src/components/interactive-image/interactive-image.scss
MarkMed: My bad. I forgot to return it to the original value. I'll fix that

MarkMed: @ncardeli those css vars works for the img container. not for the img tag

src/components/interactive-image/interactive-image.scss
MarkMed: You're right.

src/components/interactive-image/interactive-image.tsx
MarkMed: 😥

src/components/interactive-image/interactive-image.scss
ncardeli: I understand, but why should I use those properties instead of directly using `width` and `height` on the custom element?

src/components/interactive-image/interactive-image.tsx
MarkMed: For more intuitive call and edit. Perhaps the names should be changed.

`zoomFeature.over.touchBehavior` if you read that you can easily understand that _touchBehavior_ should be some function that runs when the user is over the component. Do you get it?
That is the logic. All the zoom stuff in one

src/components/interactive-image/interactive-image.scss
MarkMed: If we just pass `width` and `height` as attributes... The component read it well? 😯

MarkMed: I try it and it doesn't work. Do you say add it as properties?

MarkMed: Values fixed in da6f90c

If is desirable to add `width` and `height` as properties of the control and be available to set the size using those properties, ask me to add it!

src/components/interactive-image/interactive-image.scss
MarkMed: fixed in da6f90c

src/components/interactive-image/interactive-image.tsx
MarkMed: doc updated in da6f90c

src/components/interactive-image/interactive-image.scss
MarkMed: nope, that was for another solution. Those styles were deleted in da6f90c

src/components/interactive-image/interactive-image.tsx
MarkMed: ok ok, I saw the error. Stop bullying me 😥

src/components/interactive-image/interactive-image.tsx
MarkMed: I forgot to delete this part. It was part of another solution. I'll fix that

src/components/interactive-image/interactive-image.tsx
MarkMed: Fix the shame in 4d5e034 👌

src/components/interactive-image/interactive-image.tsx
MarkMed: That was fixed in 4d5e034

src/components/interactive-image/interactive-image.tsx
MarkMed: This shame was fixed in b690e51

src/components/interactive-image/interactive-image.tsx
MarkMed: I liked. I apply the changes in b690e51

src/components/interactive-image/interactive-image.tsx
MarkMed: I deleted that. It was within double condition... 🤦‍♂️
Look at 4d5e034

src/components/interactive-image/interactive-image.tsx
MarkMed: You are right. I forgot that detail. Thx for remind me it 👌]
Btw that part was deleted in d1cafe5

src/components/interactive-image/interactive-image.tsx
MarkMed: Deleted. Look at d1cafe5

src/components/interactive-image/interactive-image.tsx
MarkMed: That method is useless now. It was removed in d1cafe5

src/components/interactive-image/interactive-image.tsx
MarkMed: Let's add it in next features then!

MarkMed: removed in b764a48 !

src/components/interactive-image/interactive-image.tsx
MarkMed: Yeap, it was removed now. Look at commit > b764a48

src/components/interactive-image/interactive-image.tsx
ncardeli: Jaja, the heart was because I sincerely liked that you used a combination of `object-position` and `transform`. I wasn't being sarcastic. 👍

src/components/interactive-image/interactive-image.tsx
MarkMed: look changes of commit 297a6f1

src/components/interactive-image/interactive-image.scss
MarkMed: I applied your recommendation!
Look at the commit 0044d2e

MarkMed:
Next feature: _Outside zoom_

PR #120, opened 2020-07-16
Add/suggest select

MarkMed:
Suggest feature to Select Control has been created.


MarkMed:
@ncardeli You think that it should have different generated Id when ```suggest``` attr is enabled?
Eg:
```suggest = true``` > ```id="gx-data-list-auto-id-n"```

File: src/components/renders/bootstrap/select/select-render.tsx [hunk 1]
```diff
@@ -79,16 +79,28 @@ export class SelectRender implements Renderer {
...
+        ? [
+            <input list={this.selectId}></input>,
+
+            <datalist id={this.selectId}>
+              {this.options.map(({ disabled, innerText, selected, value }) => (
+                <option disabled={disabled} selected={selected} value={value}>
+                  {innerText}
+                </option>
+              ))}
+            </datalist>
```
ncardeli, line 92: We should also include `gx-bootstrap` (at least until we use the new design). We should also set the `form-input` class to the `<input>` element.

src/components/renders/bootstrap/select/select-render.tsx
```diff
@@ -79,16 +79,28 @@ export class SelectRender implements Renderer {
...
-              {innerText}
-            </option>
-          ))}
-        </select>
-      ];
+      return this.component.suggest
+        ? [
+            <input list={this.selectId}></input>,
+
+            <datalist id={this.selectId}>
```
ncardeli, line 86: I think the `datalist` should have its own id, something like:

```
      const dataListId = `${this.selectId}_datalist`;
```

And then you could have something like this:

```
            <input list={dataListId}></input>,
            <datalist id={dataListId}>
```

src/components/renders/bootstrap/select/select.e2e.ts
```diff
@@ -39,4 +39,11 @@ describe("gx-select", () => {
     await page.waitForChanges();
     expect(await page.find("select")).toHaveClass("foo-class");
   });
+
+  it("should set the input with suggest list if suggest attr is enabled", async () => {
+    const input = await element.find("input[list='gx-select-auto-id-0']");
```
ncardeli, line 44: I think that using the `id` makes this test too brittle and that the id is an internal implementation detail.
We could achieve the same result without referencing the `id` and by checking that the `list` attribute of the `input` element and the `id` of the `datalist` element are equal.

src/components/select/select.tsx
```diff
@@ -57,6 +57,11 @@ export class Select implements FormComponent {
    */
   @Prop() readonly readonly: boolean;
 
+  /**
+   * Enable a text imput with suggest values.
```
ncardeli, line 61:
```suggestion
   * Render a text input showing a list of suggested elements.
```

src/components/renders/bootstrap/select/select-render.tsx
MarkMed: Bootstrap...
Ok ok, I will add it! 👌

src/components/renders/bootstrap/select/select-render.tsx
MarkMed: Yeap, that I asked you in previous comments. I will add that change 😉

src/components/select/select.tsx
MarkMed: Ready! Check the commit 15714ff to see the change

src/components/renders/bootstrap/select/select.e2e.ts
MarkMed: I updated that e2e, check it in 5709850

src/components/renders/bootstrap/select/select-render.tsx
MarkMed: Setting a variable for the string of `id` and evaluate if `suggest` is enabled and then set it... that way is complex.
I prefer to simply add the "___datalist_" to the `id`. This: `${this.selectId}_datalist`
Checkout the commit f25ad5a to see that simply change

src/components/renders/bootstrap/select/select-render.tsx
MarkMed: Bootstrap tag added in commit f25ad5a
("I forgot to add a description for this change in the last commit 😅")

src/components/renders/bootstrap/select/select-render.tsx
```diff
@@ -79,16 +79,29 @@ export class SelectRender implements Renderer {
...
-            </option>
-          ))}
-        </select>
-      ];
+      return this.component.suggest
+        ? [
+            <gx-bootstrap />,
+            <input list={`${this.selectId}__datalist`}></input>,
+
+            <datalist id={`${this.selectId}__datalist`}>
```
ncardeli, line 87: Imagine that in the future we want to change the suffix, we would have to change it in two places. What can we do to avoid having that duplication?

MarkMed: You are right, it's better to save that string in a variable and just edit it in one place. DRY principle.
Check 0df5ffc commit!

src/components/renders/bootstrap/select/select-render.tsx
```diff
@@ -78,13 +79,16 @@ export class SelectRender implements Renderer {
           select.value = this.component.value;
         }
       };
+      if (this.component.suggest) {
+        this.datalistId = `${this.selectId}__datalist`;
```
ncardeli, line 83: Just an observation: instead of a property, this could have been a local variable inside the `render` method as it's not used outside of the method.

MarkMed:
I applied your recommendation! I will merge now to make it available!

PR #127, opened 2020-07-31
Quick fix of Checkbox Control

MarkMed:
The error is caused by the emission of two events with the same name.
The `input` element (inside the checkbox render) listens to the `input` custom event and the native `input` event, taking the value of the last event.

Fix solution: prevent emittion (propagation) of "native" `input` event.
> image

### DIFF
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -103,6 +103,7 @@ export class CheckBox implements FormComponent {
   }
 
   private handleChange(event: UIEvent) {
+    event.stopPropagation();
     this.checked = this.renderer.getValueFromEvent(event);
     this.updateValue();
     this.input.emit(event);


MarkMed:
To reproduce the error use the following code in the main HTML `<body>`:
```HTML
<gx-checkbox id="gxCheck" checked-value="Checked!" un-checked-value="Unchecked" checked=true></gx-checkbox>
<p id="testParag"></p>
<script>
    const parag = document.getElementById("testParag");
    document.getElementById("gxCheck").addEventListener("input", (ev)=>{
	// console.log("ev.detail.returnValue", ev.detail.returnValue)
	console.log("--------")
	console.log("ev >", ev);
	console.log("ev.target.checked >", ev.target.checked);
	console.log("ev.target.value >", ev.target.value);
	console.log("--------")
	parag.innerText = ev.target.value;
    });
</script>
```
Run this code in the this branch and in the master branch. To test with the fix and without the fix 👍

MarkMed:
Please, keep me informed about this control and how it works!

PR #129, opened 2020-08-13
Image component not honoring width when inside Flex container

ggallotti:
@ncardeli:
The problem is that when the Image Component is placed inside a Flexbox container, the Width is not honored, but fills the entire container. 

image

I think that the Image component is the only component whose size can be established via Theme with Image Width and Height (and may not fill the entire "row".) (And not inherited from the parent Cell size. )

I also considered setting the Width to the IMG, but we must resize the whole container, not only the IMG inside, because we expect that Borders are applied to the whole container. 

I've added a maxWidth attribute for this. 

@ncardeli what do you think? Better ideas are welcome. ;)

### DIFF
--- a/src/components/image/image.tsx
+++ b/src/components/image/image.tsx
@@ -178,6 +178,12 @@ export class Image
 
     const isHeightSpecified = !!this.height;
     const isWidthSpecified = !!this.width;
+    const height = isHeightSpecified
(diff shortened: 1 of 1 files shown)


ncardeli:
Sounds good to me; just verify that everything works as expected when `autoGrow=true` and `autoGrow=false`.
This component changes the way it renders radically when `autoGrow` is toggled.

PR #130, opened 2020-08-14
fix for disable state

MarkMed:
`disable` prop for button element inside gxButton control


MarkMed:
@ncardeli We must find a way to prevent that "capturing event effect". By attaching the _click_ event in gxButton, the `span` or `img` capturing the event and running the callback function.

PR #131, opened 2020-08-18
tab error fixed

MarkMed:
Now, don't need to interact directly with tab control, you can change the tab with btns or another extern method (with Btns as @ag-silva did in his test case)

### DIFF
--- a/src/components/renders/bootstrap/tab-caption/tab-caption-render.tsx
+++ b/src/components/renders/bootstrap/tab-caption/tab-caption-render.tsx
@@ -54,6 +54,7 @@ export class TabCaptionRender implements Renderer {
 
   private clickHandler(event: UIEvent) {
     event.preventDefault();
+    this.component.selected = true;
     this.component.tabSelect.emit(event);
   }
 }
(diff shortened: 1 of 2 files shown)


src/components/renders/bootstrap/tab-caption/tab-caption-render.tsx
```diff
@@ -54,6 +54,7 @@ export class TabCaptionRender implements Renderer {
 
   private clickHandler(event: UIEvent) {
     event.preventDefault();
+    this.component.selected = true;
```
ncardeli, line 57: I'm worried that, by changing the property here and emitting the `tabSelect` event in the next line, the event will get fired twice when the tab page is clicked. Can you verify it please?

ncardeli: Now we are emitting the `tabSelect` event in two different places. How can we fire it in only one place, so we can have only one source of truth to decide when to emit the event?
We already have a `@Watch` on the `selected` property, so maybe we could take advantage of that.

MarkMed: 🤔

MarkMed:
@ncardeli in the `clickHandler` method should be included the `this.component.selected = true;` line. To change the `select` attribute, and then, the `Watch` decorator will emit the `tabSelect`

PR #134, opened 2020-08-26
Fix/button

MarkMed:
`Listen` decorator added to prevent click capturing

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -90,8 +90,8 @@ export class Button
 
   @Listen("click", { capture: true })
   private handleClick(event: UIEvent) {
+    event.stopPropagation();
     if (this.disabled) {
-      event.stopPropagation();
       return;
     }
 
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -1,15 +1,6 @@
 import { ButtonRender } from "../renders/bootstrap/button/button-render";
+import { Component, Element, Prop, Listen, h } from "@stencil/core";
 import {
-  Component,
-  Element,
-  Event,
-  EventEmitter,
-  Prop,
-  Listen,
-  h
-} from "@stencil/core";
-import {
-  ClickableComponent,
   Component as GxComponent,
   DisableableComponent,
   VisibilityComponent
@@ -21,11 +12,7 @@ import {
   tag: "gx-button"
 })
 export class Button
-  implements
-    GxComponent,
-    ClickableComponent,
-    DisableableComponent,
-    VisibilityComponent {
+  implements GxComponent, DisableableComponent, VisibilityComponent {
   constructor() {
     this.renderer = new ButtonRender(this, {
       handleClick: this.handleClick.bind(this)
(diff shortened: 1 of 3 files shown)


src/components/button/button.tsx
```diff
@@ -87,19 +88,21 @@ export class Button
    */
   @Prop() readonly size: "large" | "normal" | "small" = "normal";
 
-  /**
-   * Emitted when the element is clicked.
-   */
-  @Event() gxClick: EventEmitter;
-
+  @Listen("click", { capture: true })
```
ncardeli, line 91: I think that with this approach, when `disabled=false`, the `gxClick` event will fire twice. Once in the capture phase and once in the bubbling phase. Is that right?

Additionally, once the click is cancelled when `disabled` is `true`, does having the `gxClick` event make sense?

MarkMed: Your approach is accurate. I tested and it emits `gxClick` twice. 
Look at the next commit. I cancel the event propagation, and now, event target is emitted from gxButton.

MarkMed:
@ncardeli, in a0c603e: The `gxClick` is emitted only from `gxButton` and only if it's available (`disable="false"`)

To make a quick test, use the following code: 
```HTML
<body>
    <h1>Test page</h1>
	<br>
	<gx-button id="gxBtn" disabled="false">
		This a DISABLED button
	</gx-button>
	<p id="wea"></p>
	<script>
		let counterStrike = 0;
		document.getElementById("gxBtn").addEventListener("click", (ev)=>{
			console.log("this event")
			console.log(ev);
			console.log(ev.target);
			counterStrike++
			document.getElementById("wea").innerText=counterStrike;
		});
		
		window.addEventListener("gxClick", (ev)=>{
			console.log("GxClick!", ev.target);
		});
	</script>
  </body>
```

src/components/button/button.tsx
ncardeli: It looks better now! However, do we need the `gxClick` event now that the `click` event is canceled when `disabled` is `true`?

MarkMed: @ncardeli removing `gxClick` event, there is a conflict with the interface of `ClickableComponent`

ncardeli: Yes, exactly. Feel free to remove that interface from the component.

MarkMed:
With the `@Listen()` decorator added, I prevent the `click` capturing, but... if I add a listener, the callback function doesn't run.

Example:
``` javascript
gxButton.addEventListener("click", (ev)=>{console.log("btnClicked!")});
```
If I click the button, it doesn't log the "btnClicked!". It runs the `handleClick` from gxButton

src/components/button/button.tsx
```diff
@@ -87,19 +88,21 @@ export class Button
...
   private handleClick(event: UIEvent) {
+    event.stopPropagation();
```
ncardeli, line 93: Shouldn't we stop propagation only when `disabled` is `true`?

MarkMed: I suspect when control is enabled and clicked, it will bubbling/capturing...
Before it was that way... the stop propagation only when `disabled` is `true`... I don't remember why I changed it.
Btw, I will try putting the `stopPropagation`  inside a condition if `disabled` is `true`.

MarkMed: @ncardeli putting the `stopPropagation` inside a condition if  `disabled` is `true`, the bubbling/capturing appears!
image

MarkMed:
@ncardeli to summarize the new fix:

It prevents the click bubbling/capturing when the button is disabled. But, if the button is enabled and you click it, the event will bubbling/capturing.

Btw, the main issue is already fixed!

ncardeli (approved):
💯

PR #141, opened 2020-09-02
Tab Page was not taking 100% of available height.

ggallotti:
Tab Page content was not taking 100% of available space inside Tab Control.


ggallotti:
Will be addressed by @ncardeli in new tab control version.

PR #156, opened 2020-09-10
Add/gx map circle

fedes1987:
- `gx-map-circle` control was added. It allows create circles in the `gx-map`.

- At this moment, the control does not support styling.


src/components/map-circle/map-circle.tsx
```diff
@@ -0,0 +1,77 @@
...
+    this.gxMapCircleDidLoad.emit(this.circleInstance);
+  }
+
+  componentDidUnload() {
+    this.gxMapCircleDeleted.emit(this.circleInstance);
+  }
+
+  render() {
+    return (
+      <div class="popup-data-container">
```
ncardeli, line 72: I think this class is not used

src/components/map/map.tsx
```diff
@@ -226,6 +245,22 @@ export class Map implements GxComponent {
...
+    ) {
+      i++;
+    }
+    if (i <= this.circleList.length) {
+      this.circleList.splice(i, 1);
+    } else {
+      console.warn("There was an error in the circle list!");
+    }
+  }
+
```
ncardeli, line 263: It would be great if we could reuse this logic for the different types of shapes so that we can avoid code repetition. What do you think?
Feel free to address this in a future PR

ncardeli (approved):
Great work, thank you!

src/components/map-circle/map-circle.tsx
fedes1987: Done. Addressed in 451a051

ncardeli (approved):

dangmarm (approved):

PR #158, opened 2020-09-16
Fix/gx select

MarkMed:
Bug > `gxSelect` value and selected option no matching
Fix > Improve value setting when getting `gxSelectOption`s


src/components/select-option/select-option.tsx
```diff
@@ -97,7 +97,15 @@ export class SelectOption implements GxComponent, DisableableComponent {
...
     return (
-      <Host aria-hidden="true" hidden={true}>
+      <Host
+        aria-hidden="true"
+        hidden={true}
+        selected={
+          this.element.parentElement.getAttribute("value") === this.value
+            ? "true"
+            : "false"
+        }
```
ncardeli, line 107: Why are we setting one of the component's properties here?
It doesn't seem right. `Host` allows us to set standard attributes of the host element (like `style`, `class`, `role`, or `aria-*`), but setting properties of the own component doesn't seem a good practice. Instead, we should just set the property `this.selected = someValue`. If you want the property value to be reflected as an attribute, we should use the  `reflect` option of the `Prop()` decorator (more information about [reflect](https://stenciljs.com/docs/properties#reflect-properties-values-to-attributes)).

src/components/select/select.tsx
```diff
@@ -59,10 +59,14 @@ export class Select implements FormComponent {
 
   /**
    * Render a text input showing a list of suggested elements.
-
    */
   @Prop() suggest: boolean;
 
+  /**
+   * Specifies a short hint that describes the expected value of the field.
```
ncardeli, line 66:
```suggestion
   * Text that appears in the form control when it has no value set.
```

src/components/select/select.tsx
```diff
@@ -77,12 +81,14 @@ export class Select implements FormComponent {
...
-        disabled: option.getAttribute("disabled") !== null,
-        innerText: option.innerText,
-        selected: option.getAttribute("selected") !== null,
-        value: option.value
-      })
+      (option: any) => {
+        return {
+          disabled: option.getAttribute("disabled") === "true" ? true : false,
+          innerText: option.innerText,
+          selected: option.getAttribute("selected") === "true" ? true : false,
```
ncardeli, line 88: Maybe this is a good opportunity to use the properties instead of the attributes:
```suggestion
          disabled: option.disabled,
          innerText: option.innerText,
          selected: option.selected,
```

ncardeli (changes requested):
Looks good!
I left some comments for you to consider.

src/components/select/select.tsx
MarkMed: That mapping has been created by you. There, you do a `querySelectorAll` to get all children. And those children are HTMLElements. I cannot access to those properties like a component...
Btw, I will try to do that!

src/components/select/select.tsx
MarkMed: I take the description from w3schools hehe.

src/components/select/select.tsx
ncardeli: Yes, I remember I did, but everything can be improved 😄

src/components/select-option/select-option.tsx
MarkMed: By editing the `this.selected = true/false` will provoke another render execution, and then, another emittion of `gxDidLoad` event.
Setting as an attribute, the `gxSelect` will read the `gxSelectOption` with that attribute, and set it to the intern `option` of  intern `input select`.
Instead of setting the `selected` attribute manually, It's set automatically checking the `value` of `gxSelect`.

src/components/select/select.tsx
ncardeli: Yes, I thought it wasn't clearly explaining what the attribute thas (placeholder text doesn't need to be short nor has to show the expected value, even though it's often used that way) so I looked the one I proposed on MDN, which, BTW, is a much better quality and trustworthy resource than w3schools.

src/components/select-option/select-option.tsx
ncardeli: Please bear in mind that `componentDidLoad` fires only after the first render, so no problem with that.

I think we could explore here setting `selected` from the parent component, that knows its value and "sees" the child components.

src/components/select/select.tsx
MarkMed: Edited document in 7d60907

src/components/select/select.tsx
MarkMed: I realized you can access it! What weirdo...
Changes applied in bf88e1d

src/components/select-option/select-option.tsx
MarkMed: Checkout changes about this in commit 7a92b8b

src/components/select/select.tsx
```diff
@@ -139,23 +145,18 @@ export class Select implements FormComponent {
...
+    if (this.value) {
+      if (this.value === option.value) {
+        option.selected = true;
+      } else {
+        option.selected = false;
+      }
+    } else {
+      if (option.selected) {
+        this.value = option.value;
+      }
```
ncardeli, line 157: Why did we remove all the comments? I think they were relevant.

MarkMed: I thought it was pretty intuitive...
do you say to add them again?

ncardeli:
Yes please :)

MarkMed:
@ncardeli please, check the comments and tell me what do you think!

src/components/select/select.tsx
```diff
@@ -129,15 +143,32 @@ export class Select implements FormComponent {
...
+      // if the select does not have a value
+      // let's look for options initialized as checked
       if (option.selected) {
+        // this option was initialized as checked,
+        // so let's set the select's value
+        // equals to the checked option value
         this.value = option.value;
       }
+      // if no option is checked,
+      // all is unchecked and the select value
```
MarkMed, line 170: perhaps here I can say something better... @ncardeli what do you think to comment here instead?

MarkMed: @ncardeli 
maybe the comment could be something like:
```
      // If there is no option checked 
      // and no value was set in the select,
      // it will keep undefined until any 
      // change or checked option
```

src/components/select/select.tsx
MarkMed: @ncardeli comments are back!
Check them in f6a05fe

src/components/select/select.tsx
ncardeli: Yes, looks good

MarkMed: this change was applied in f7924f0

src/components/select/select.tsx
```diff
@@ -194,14 +210,14 @@ export class Select implements FormComponent {
...
-  private setDisabled() {
-    this.options.forEach(option => {
-      option.disabled = this.disabled;
-    });
-  }
+  //   private setDisabled() {
+  //     this.options.forEach(option => {
+  //       option.disabled = this.disabled;
+  //     });
+  //   }
```
ncardeli, line 217: Let's remove the commented code

src/components/select/select.tsx
```diff
@@ -194,14 +210,14 @@ export class Select implements FormComponent {
...
 
   componentDidLoad() {
-    this.setDisabled();
+    // this.setDisabled();
```
ncardeli, line 220: Same here

ncardeli (approved):
Approving but remember to remove the commented code before merging.

src/components/select/select.tsx
MarkMed: This is was asked to you in commit f8fff15
If I should remove that method or not. That method provokes the issue with the disable/enable option.
Should I remove the comment or remove the whole method?

ncardeli: I mean removing the block of commented code, of course. Not uncommenting it.

MarkMed: @ncardeli I think it should be better if we leave the method commented, maybe it will be necessary in the future. Do you suggest to remove it (delete it from the code)?

ncardeli: Yes, if we want to recover it, we have the history that Git stores for us.

MarkMed: Ok, I will remove it now

src/components/select/select.tsx
MarkMed: I'll remove this too

MarkMed: Removed in commit 3feecfc

PR #168, opened 2020-10-01
triggerText now is displayed if slot is not setted

MarkMed:
When `show-trigger = true`  you can set the `trigger-text` with some text to show within the button.
The bug was an error where `trigger-text` never was within the button. 

Now, if  `show-trigger = true` and you don't set a `slot` element, the `trigger-text` will be displayed.
If you set a `slot` element,  the `trigger-text` value only will be as an aria-attr value, and the button will the content setted in the `slot` element.

### DIFF
--- a/src/components/renders/bootstrap/edit/edit-render.tsx
+++ b/src/components/renders/bootstrap/edit/edit-render.tsx
@@ -112,6 +112,9 @@ export class EditRender implements Renderer {
       const input = <input {...attris} type={edit.type} value={edit.value} />;
 
       if (edit.showTrigger) {
+        const existSlotContent = edit.element.querySelector(
+          "[slot='trigger-content']"
+        );
         editableElement = (
           <div class="input-group" hidden={edit.readonly}>
             {input}
@@ -123,7 +126,9 @@ export class EditRender implements Renderer {
                 disabled={edit.disabled}
                 aria-label={edit.triggerText}
               >
-                {slots.triggerContent}
+                {existSlotContent !== null
+                  ? slots.triggerContent
+                  : edit.triggerText}
               </button>
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


src/components/renders/bootstrap/edit/edit-render.tsx
```diff
@@ -69,6 +69,35 @@ export class EditRender implements Renderer {
     event.stopPropagation();
   }
 
+  setReadonlyContent(component, initialContent) {
+    let content = initialContent;
+    console.log(content);
```
ncardeli, line 74: 👀

src/components/renders/bootstrap/edit/edit-render.tsx
```diff
@@ -69,6 +69,35 @@ export class EditRender implements Renderer {
     event.stopPropagation();
   }
 
+  setReadonlyContent(component, initialContent) {
```
ncardeli, line 72: More than setting, I believe that with this method we are getting. What do you think?

ncardeli (changes requested):
Nice work here. Found a `console.log` (ups) and left a small comment on a method name

src/components/renders/bootstrap/edit/edit-render.tsx
MarkMed: That method takes the content and returns the content with a new format when the component is readonly mode and it has a specific type. 🤔
What name do you suggest?

ncardeli: If it returns something, `getReadonlyContent` sounds more accurate.

src/components/renders/bootstrap/edit/edit-render.tsx
MarkMed: ups... removed in 534eb88

src/components/renders/bootstrap/edit/edit-render.tsx
MarkMed: Renamed! Look in 0be53e5

ncardeli (approved):

PR #180, opened 2020-10-14
gxClick to click

MarkMed:
setting the new click handling logic to the controls.
The new approach has been applied on:
 - gx-image
 - gx-textblock
 - gx-button
 - gx-table

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -16,7 +16,7 @@ export class Button
   implements GxComponent, DisableableComponent, VisibilityComponent {
   constructor() {
     this.renderer = new ButtonRender(this, {
-      handleClick: this.handleClick.bind(this)
+      handleClick: !this.disabled ? this.handleClick.bind(this) : null
     });
   }
 
@@ -77,7 +77,8 @@ export class Button
   @Prop() readonly size: "large" | "normal" | "small" = "normal";
 
   @Listen("click", { capture: true })
-  private handleClick(event: UIEvent) {
+  handleClick(event: UIEvent) {
+    event.preventDefault();
     if (this.disabled) {
       event.stopPropagation();
       return;
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -16,7 +16,7 @@ export class Button
   implements GxComponent, DisableableComponent, VisibilityComponent {
   constructor() {
     this.renderer = new ButtonRender(this, {
-      handleClick: !this.disabled ? this.handleClick.bind(this) : null
+      handleClick: this.handleClick.bind(this)
     });
   }
 
@@ -78,7 +78,6 @@ export class Button
 
   @Listen("click", { capture: true })
   handleClick(event: UIEvent) {
-    event.preventDefault();
     if (this.disabled) {
       event.stopPropagation();
       return;
--- a/src/components/image/image.tsx
+++ b/src/components/image/image.tsx
@@ -114,7 +114,7 @@ export class Image
   @Prop({ mutable: true }) width: string;
 
   @Listen("click", { capture: true })
-  private handleClick(event: UIEvent) {
+  handleClick(event: UIEvent) {
     event.preventDefault();
     if (this.disabled) {
       event.stopPropagation();
--- a/src/components/image/image.tsx
+++ b/src/components/image/image.tsx
@@ -1,15 +1,6 @@
-import {
-  Component,
-  Element,
-  Event,
-  EventEmitter,
-  Prop,
-  h,
-  Host
-} from "@stencil/core";
+import { Component, Element, Prop, h, Listen, Host } from "@stencil/core";
 import lazySizes from "lazysizes";
 import {
-  ClickableComponent,
   Component as GxComponent,
   DisableableComponent,
   VisibilityComponent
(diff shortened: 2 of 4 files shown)


src/components/button/button.tsx
```diff
@@ -16,7 +16,7 @@ export class Button
   implements GxComponent, DisableableComponent, VisibilityComponent {
   constructor() {
     this.renderer = new ButtonRender(this, {
-      handleClick: this.handleClick.bind(this)
+      handleClick: !this.disabled ? this.handleClick.bind(this) : null
```
ncardeli, line 19: What happens if `disabled` is changed in runtime?

src/components/button/button.tsx
```diff
@@ -77,7 +77,8 @@ export class Button
   @Prop() readonly size: "large" | "normal" | "small" = "normal";
 
   @Listen("click", { capture: true })
-  private handleClick(event: UIEvent) {
+  handleClick(event: UIEvent) {
+    event.preventDefault();
```
ncardeli, line 81: Why do we need to prevent the default behavior?

src/components/image/image.tsx
```diff
@@ -126,18 +113,13 @@ export class Image
...
 
-  /**
-   * Emitted when the element is clicked.
-   */
-  @Event() gxClick: EventEmitter;
-
-  private handleClick(event: UIEvent) {
+  @Listen("click", { capture: true })
+  handleClick(event: UIEvent) {
+    event.preventDefault();
```
ncardeli, line 118: Why are we preventing default?

ncardeli (changes requested):
This is a solid start. I left some comments.

src/components/button/button.tsx
MarkMed: I include the preventDefault to prevent unexpected behavior... Do you suggest to remove that line?

src/components/image/image.tsx
MarkMed: I include the preventDefault to prevent unexpected behavior... Do you suggest to remove that line?

ncardeli: I'd remove it if it's just in case.

src/components/button/button.tsx
MarkMed: mmm... I thought it will be great but I didn't think in change the `disabled` in runtime. I'll remove that change

src/components/button/button.tsx
MarkMed: removed! Check 09ae8dc

src/components/image/image.tsx
MarkMed: removed! Check 09ae8dc

src/components/button/button.tsx
MarkMed: Fixed in 09ae8dc

src/components/table/table.tsx
```diff
@@ -87,6 +88,15 @@ export class Table
    */
   @Event() swipeLeft: EventEmitter;
 
+  @Listen("click", { capture: true })
+  handleClick(event: UIEvent) {
+    event.preventDefault();
```
ncardeli, line 93: Why are we preventing the default behavior here?

src/components/button/button.tsx
```diff
@@ -77,7 +77,7 @@ export class Button
   @Prop() readonly size: "large" | "normal" | "small" = "normal";
 
   @Listen("click", { capture: true })
-  private handleClick(event: UIEvent) {
+  handleClick(event: UIEvent) {
```
ncardeli, line 80: Can we keep the method private?

src/components/image/image.tsx
```diff
@@ -126,18 +113,12 @@ export class Image
...
   @Prop({ mutable: true }) width: string;
 
-  /**
-   * Emitted when the element is clicked.
-   */
-  @Event() gxClick: EventEmitter;
-
-  private handleClick(event: UIEvent) {
+  @Listen("click", { capture: true })
+  handleClick(event: UIEvent) {
```
ncardeli, line 117: Can we keep this method private?

MarkMed: it isn't private... it's a handle for the event assigned ("clicked").
Can it be the handler and be private at same time?

ncardeli: Yes, it can.

src/components/button/button.tsx
MarkMed: it isn't private... it's a handle for the event assigned ("clicked").
Can it be the handler and be private at same time?

src/components/table/table.tsx
MarkMed: Same to previous logic, prevent possible default functions. I will remove that.

src/components/image/image.tsx
MarkMed: If I set the handler as  `private`, it takes that as a method...

src/components/button/button.tsx
ncardeli: Ok, I understood the issue here. The `@Listen` decorator hides the fact that the `handleClick` method is actually used, and the TS compiler believes that it's not called, thus the warning.
It's ok the way you coded it 👍

src/components/table/table.tsx
ncardeli: I'm accepting this PR, but please don't forget to remove the `preventDefault` call

ncardeli (approved):

src/components/table/table.tsx
MarkMed: I will commit this last change and merge the PR

PR #181, opened 2020-10-20
Fix/gx radio group option

MarkMed:
### Issue
When you assign a `gxRadioGroup` as `read-only`, the radio-buttons still available and clickable.
Another issue was about the label position, but this bug was in `gxFormField`. You assign the `label-position` but the label only show properly when `label-position = "left"`

### Fixes
Unlike gxEdit, if we take out the input (radiobutton) and left only the label, it will be a bit confusing. That is why I decided to show as "disabled" when readioButton is `read-only`
And for the _label position_ issue, I decided to use the `flex-direction` to order the label and the radioButtons

### Additional Idea
Perhaps we should set a CSS class in the `.sass` file and use that class instead of inline style. What do you think?

### DIFF
--- a/src/components/radio-group/radio-group.tsx
+++ b/src/components/radio-group/radio-group.tsx
@@ -59,6 +59,13 @@ export class RadioGroup
   @Prop() readonly name: string;
 
   /**
+   * This attribute indicates that the user cannot modify the value of the control.
+   * Same as [readonly](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-readonly)
+   * attribute for `input` elements.
+   */
+  @Prop() readonly readonly: boolean;
+
+  /**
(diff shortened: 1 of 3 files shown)


dangmarm:
I see that there are defined in the .scss classes label-left, label-right, etc. The flex-direction property should not be with its corresponding value in the corresponding class instead of defining it in the element's style?

MarkMed:
I put the in-line style for a quick Fix. I will implement that fix in the style classes

MarkMed:
@dangmarm check out the new changes in 5e8ca06
U will notice that I created new CSS classes in order to set the flex-direction of the parent element instead of doing in-line.

And then I will add new changes to fix the behavior of `radioGroup` when it is in `read-only` mode. That will be discussed tomorrow.

dangmarm (approved):

PR #182, opened 2020-10-21
fixing UI issue (size)

MarkMed:
### Issue
`gxEdit` and `gxPasswordEdit` within `gxTable` has different heights

### Fix
Set the 100% of the height of the inner `gxEdit` of `gxPasswordEdit`

#### Additional Idea
_None_


dangmarm (approved):
👍

PR #184, opened 2020-11-19
Add/gx gauge

MarkMed:
Early PR of Gauge control.
The idea of this PR is to look at the code, run test cases, talk about implementation, talk about new ideas, fixes, etc.

This is a complex control where SVG technology is required.


MarkMed:
## Current tasks to be done:
+ Add **labels** in `circle` mode

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
+@import "../common/_base";
+
+gx-gauge {
+  /**
+  * @prop --component-height:
+   * Set the height or the control. *Must be set before render the control*
+   * (400px by default)
+  */
+  --component-height: 400px;
```
ncardeli, line 9: Can we use `height` to set the component's height instead of a custom property?

ncardeli: See here: https://github.com/genexuslabs/web-controls-library/commit/0044d2eecc1d975c3dd8191702115ae1a109108f

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
...
+   * (gray by default)
+  */
+  --center-circle-text-color: rgba(0, 0, 0, 0.7);
+
+  /**
+  * @prop --marker-backgorund:
+   * Define the background color of the marker.
+   * (1px by default)
+  */
+  --marker-backgorund: rgba(0, 0, 0, 0.7);
```
ncardeli, line 23: Is this property intended just for setting a background color?
If so, I think it should be named `--marker-background-color`.
BTW, it seems there are other properties in the same situation.

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
...
+   * Set the color of minimum value display text.
+   * (rgba(255, 255, 255, 0.8) by default)
+  */
+  --min-value-display-color: rgba(255, 255, 255, 0.8);
+
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  flex-wrap: wrap;
+  /* justify-content: center;*/
```
ncardeli, line 57: Please don't forget to remove commented code

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
...
+  height: var(--component-height);
+
+  .gaugeContainerLine {
+    // margin-bottom: 20px;
+    width: 98%;
```
ncardeli, line 62: Why 98%? Will this work in very wide spaces as in very small spaces?

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
...
+    height: auto;
+    // display: flex;
+    // align-items: center;
+    position: absolute;
+    border-radius: 26px;
```
ncardeli, line 67: Why `26px`? Maybe we should allow customizing this property with a CSS var

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
+@import "../common/_base";
+
+gx-gauge {
```
ncardeli, line 3: I see there are multiple different values for the transition property? Shouldn't these values be the same?
And in that case, can we avoid code repetition?

ncardeli, line 3: Great to see the the CSS vars documented!

src/components/gauge/gauge.scss
```diff
@@ -0,0 +1,222 @@
...
+      border-radius: 50%;
+      z-index: 0;
+      position: absolute;
+    }
+    svg {
+      z-index: -1;
+    }
+    .gauge {
+      border-radius: 50%;
+      font-size: 1.1rem;
```
ncardeli, line 198: I'd avoid setting the font size and allow it to cascade to allow for more customization. What do you think?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+  /**
+   * The `gxGaugeDidLoad` event is triggered when component has been rendered completely.
+   */
+  @Event() gxGaugeDidLoad: EventEmitter;
+
+  /**
+   * Property of type Style.
+   * Define if shadow will display or not. Default is disabled.
+   */
+  @Prop() styleShadow = false;
```
ncardeli, line 31: How about just "shadow"?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+export class Gauge implements GxComponent {
+  @Element() element: HTMLGxGaugeElement;
+
+  /**
+   * The `gxGaugeDidLoad` event is triggered when component has been rendered completely.
+   */
+  @Event() gxGaugeDidLoad: EventEmitter;
+
+  /**
+   * Property of type Style.
```
ncardeli, line 28: Not sure what this means. I'd remove it, the next line correctly expresses what the property is doing.

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+@Component({
+  shadow: false,
+  styleUrl: "gauge.scss",
+  tag: "gx-gauge"
+})
+export class Gauge implements GxComponent {
+  @Element() element: HTMLGxGaugeElement;
+
+  /**
+   * The `gxGaugeDidLoad` event is triggered when component has been rendered completely.
```
ncardeli, line 23:
```suggestion
   * The `gxGaugeDidLoad` event is triggered when the component has been rendered completely.
```

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+  @Prop() value: number;
+
+  /**
+   * Allows specify the width of the circumference _(When gauge is circle type)_ or the width of the bar _(When gauge is Line type)_ in % relative the component size.
+   *
+   */
+  @Prop() thickness = 10;
+
+  /**
+   * The maximum value of the gauge. _Calculated by Default_
```
ncardeli, line 63: How is it calculated?

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+          </div>
+        ) : (
+          ""
+        )}
+      </div>
+    );
+  }
+
+  render() {
+    console.log("this.maxValue", this.maxValue);
```
ncardeli, line 351: 👀

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+      const index = this.rangesChildren.findIndex(
+        elementFinding => elementFinding === childRange
+      );
+      this.rangesChildren.splice(index, 1, childRange);
+      this.maxValueAux = 0;
+      for (const childInstance of this.rangesChildren) {
+        this.maxValueAux += childInstance.amount;
+      }
+    });
+  }
```
ncardeli, line 98: Please consider this alternate approach for keeping a record of the component's children: https://github.com/genexuslabs/web-controls-library/blob/51e0895e37a3086e4b63871db70a37fac2008f87/src/components/navbar/navbar.tsx#L103

src/components/gauge/gauge.tsx
```diff
@@ -0,0 +1,373 @@
...
+
+  @Listen("gxGaugeRangeDidLoad")
+  onGaugeRangeDidLoad({ detail: childRange }) {
+    this.rangesChildren = [...this.rangesChildren, childRange];
+    this.maxValueAux += childRange.amount;
+    childRange.element.addEventListener("gxGaugeRangeDidUnload", () => {
+      this.rangesChildren = this.rangesChildren.filter(
+        elementToSave => elementToSave != childRange
+      );
+      console.log(this.rangesChildren);
```
ncardeli, line 85: 👀

ncardeli (commented):
Left some initial comments for you to consider.
Thanks for the hard work!

MarkMed:
e2e test must be updated

src/components/gauge/gauge.tsx
MarkMed: all consoleLogs has been removed in ae8d049

src/components/gauge/gauge.tsx
MarkMed: removed in ae8d049

src/components/gauge/gauge.tsx
MarkMed: I'll add a short explanation about the calculation

src/components/gauge/gauge.scss
MarkMed: No, we can't. The component needs a height with a specific value to calculate the SVG/Line sizes.
Wht do u have in mind?

ncardeli: I think we can because at line 58 the value of this CSS variable is directly assigned to the `height` property.
So it would make perfect sense to have this:

```suggestion
  height: 400px;
```

src/components/gauge/gauge.scss
MarkMed: 98% of width because it needs the be with `position: absolute`. And if we set 100% of the width, it will scape from window (and put the x-scroll)

src/components/gauge/gauge.scss
MarkMed: It makes sense. I liked the idea! See new changes in 508735d

src/components/gauge/gauge.scss
MarkMed: Comments have been removed in 508735d

src/components/gauge/gauge.scss
MarkMed: Yes! I'll rename some vars

src/components/gauge/gauge.scss
MarkMed: Mmm... set a default `height`, and then change it if somebody wants a different value.
E.g.: 
``` html
   <style>
      gx-gauge.smallGauge{
          height: 200px;
      }
   </style>
   <gx-gauge class="small-gauge">
      ...
   </gx-gauge>
```
I see it as very very useful. A default height if the developer doesn't specify it. And, if the developer set a `height`, the control uses that height.
I ran that approach with the custom CSS var... but it will be better if we can remove complexity!
That is what are you suggesting?

src/components/gauge/gauge.scss
MarkMed: Check out the new names in commit 64c6713 !

src/components/gauge/gauge.scss
MarkMed: Mmm yes. You're right.
I thought it was a good idea to set a default font-size and remove responsibility of the developer.
I will fix that

src/components/gauge/gauge.tsx
MarkMed: I think that comes from a previous feature/version... I will check that

src/components/gauge/gauge.scss
MarkMed: I apply your suggestion in commit c2e09ed

src/components/gauge/gauge.scss
MarkMed: Check new changes about font-size in commit f67e266

src/components/gauge/gauge.tsx
MarkMed: The purpose of that line is to explain the type of impact of that property. That property makes visuals/styling changes.
BTW, that is from a previous version of the control. Removed in 89570cc

src/components/gauge/gauge.tsx
MarkMed: that property comes from a previous version of the control. Removed in 89570cc

MarkMed:
### Purpose `index.html` file
The `index.html` file will be reset before the PR approbation. For now, the `index.html` provides a quick example of how the control works.

src/components/gauge/gauge.tsx
MarkMed: A more extended explanation was added on commit fc0cab9

src/components/gauge/gauge.tsx
MarkMed: Doc fixed in c52f1c4

src/components/gauge/gauge.scss
ncardeli: I think we could take advantage of the fact that it's absolute positioned and set `left: 0; right: 0;` to make it take all the available horizontal space. You can then tweak `left` and `right` values to make it perfectly fit its container.

The problem with `98%` is that it might work on some widths but on very small or large spaces, it might not work as expected.

src/components/gauge/gauge.tsx
ncardeli: Much better, thanks!

MarkMed:
### FIx
> Properly calc the min-max relation. For example: if `min: -100` and `max: 100` then `50% = 0`

src/components/gauge/gauge.scss
MarkMed: I applied your change request in the commit 91823b8

MarkMed:
## ToDo:
+ Check and complete _Change Requests_
+ Add **ranges labels** in `circle` mode
+ Update _e2e test_ file

### Note
+ `index.html` file will be reset before merge PR. For now, the changes applied are for testing purposes.

src/components/gauge/gauge.tsx
MarkMed: Can I add a comment in this part suggesting an upgrade with your alternate approach?
I would like to add the control to the library and that way makes it available to be used. And then make an improvement directly in this part!
Wht do you think?

ncardeli: Yes, let's do that!

src/components/gauge/gauge.scss
MarkMed: I applied this suggestion of your in 5db164b
I had to put `right: 4px` to prevent the x-scroll

src/components/gauge/gauge.tsx
```diff
@@ -111,13 +109,16 @@ export class Gauge implements GxComponent {
...
+    childRanges: Array<HTMLGxGaugeRangeElement>
+  ): HTMLElement {
     const FULL_CIRCLE_RADIO = 100 / 2;
     const svgRanges = [];
+    const labelsRanges = [];
     const ONE_PERCENT_OF_CIRCLE_DREGREE = 3.6;
     const radius = FULL_CIRCLE_RADIO - this.thickness / 2;
 
-    function renderSvgCircle(currentChild, position, component): HTMLElement {
+    function renderSvgCircle(currentChild, position, component): SVGElement {
```
ncardeli, line 121: This function doesn't seem to return SVGElement, but JSX. It's ok if you leave it without specifying a return type.
BTW, why not declare this function as a method and instead of passing the component as the third parameter, use `this`?

src/components/gauge/gauge.tsx
```diff
@@ -139,6 +140,17 @@ export class Gauge implements GxComponent {
       );
     }
 
+    function renderRangeLabels({ name, amount, color }): HTMLElement {
```
ncardeli, line 143: Similar to the previous comment, this function doesn't return an actual HTMLElement, but JSX (that's an object that represents an element, but not an actual DOM element).

src/components/gauge/gauge.tsx
```diff
@@ -148,87 +160,84 @@ export class Gauge implements GxComponent {
...
       this.rangesValuesAcumul += rangeValuePercentage;
-      svgRanges.splice(
-        0,
-        0,
-        renderSvgCircle(childRanges[i], positionInGauge, this)
-      );
+      svgRanges.push(renderSvgCircle(childRanges[i], positionInGauge, this));
+      labelsRanges.push(renderRangeLabels(childRanges[i]));
     }
+    console.log(labelsRanges);
```
ncardeli, line 167: 👀

src/components/gauge/gauge.tsx
```diff
@@ -148,87 +160,84 @@ export class Gauge implements GxComponent {
...
             ""
           )}
+          <div class="gauge">
+            {this.showValue ? (
+              <div>
+                <span class="current-value">{`${this.value}`}</span>
+                <span>{`${this.minValue}`}</span>
+                <span>{`-`}</span>
+                <span>{`${
+                  this.maxValue == undefined ? this.maxValueAux : this.maxValue
```
ncardeli, line 231:
```suggestion
                  this.maxValue === undefined ? this.maxValueAux : this.maxValue
```

ncardeli (commented):
We are getting really close here!

src/components/gauge/gauge.tsx
MarkMed: Added the comment in 5db164b

src/components/gauge/gauge.tsx
MarkMed: I will remove console.logs hehehe

src/components/gauge/gauge.tsx
MarkMed: I was planning to use the `bind()` method to change the reference of  `this`. But yes! I defined all those "sub" functions and declared it as a method of the component.
Change all this in 65fb97b

src/components/gauge/gauge.tsx
MarkMed: I understand the mistake in the concept. It isn't an HTML element from DOM, it's a JSX object that represents a future HTMLElement.
Fixed in 65fb97b

src/components/gauge/gauge.tsx
MarkMed: Removed in 65fb97b !

src/components/gauge/gauge.tsx
MarkMed: MInor change applied in 65fb97b! Check it out!
Tell me what do u think 😄

MarkMed:
## ToDo:
+ Update `e2e test` file

### Note
+ `index.html` file will be reset before merge PR. For now, the changes applied are for testing purposes.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,213 @@
...
+  /////////////////// Values test ///////////////////////
+  it("should automatically set the default minimum value", async () => {
+    await page.setContent(`
+      <gx-gauge show-value="true"></gx-gauge>
+    `);
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    const minValueDisplayer = await element.find("span.minValue");
+    expect(await element.getProperty("minValue")).toEqual(0);
+    expect(minValueDisplayer.innerHTML).toEqual("0<span></span>");
```
ncardeli, line 50: I think using `textContent` would simplify this case and allow you to remove `<span></span>` from the expected value.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,213 @@
...
+    const minValueTest = -100;
+    await page.setContent(`
+      <gx-gauge show-value="true" min-value="${minValueTest}"></gx-gauge>
+    `);
+    await page.waitForChanges();
+    element = await page.find("gx-gauge");
+    const minValueDisplayer = await element.find(
+      "div.minMaxDisplay span.minValue"
+    );
+    expect(await element.getProperty("minValue")).toEqual(minValueTest);
```
ncardeli, line 63: Checking this feels like we are testing Stencil instead of our logic.

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,213 @@
...
+    expect(minValueDisplayer.innerHTML).toEqual(`${minValueTest}<span></span>`);
+  });
```
ncardeli, line 65: This test case seems very similar to the previous one. Maybe we can have only one that verifies both the initial default value and the value set in the `minValue` property.

Something like this:

```
  it("should set the passed minimum value", async () => {
    await page.setContent(`
      <gx-gauge show-value="true"></gx-gauge>
    `);
    await page.waitForChanges();
    element = await page.find("gx-gauge");
    const minValueDisplayer = await element.find("span.minValue");
    expect(minValueDisplayer.textContent).toEqual("0");
    
    const minValueTest = -100;
    await element.setProperty("minValue", minValueTest);
    await page.waitForChanges();
    expect(minValueDisplayer.textContent).toEqual(`${minValueTest}`);
  });
```

src/components/gauge/gauge.e2e.ts
```diff
@@ -0,0 +1,213 @@
...
+  //     `);
+  //     await page.waitForChanges();
+  //     element = await page.find("gx-gauge");
+  //     await page.waitForChanges();
+  //     const range = await element.find("svg circle");
+  //     await page.waitForChanges();
+  //     expect(await element.find("svg")).toBeTruthy();
+  //     expect(range.getAttribute("stroke")).toEqual("rgba(116, 16, 216, 1)");
+  //   });
+});
```
ncardeli, line 213: What are the plans for this commented code?

ncardeli (commented):
I left some more comments.

src/components/gauge/gauge.e2e.ts
MarkMed: 🤔

src/components/gauge/gauge.e2e.ts
MarkMed: I will remove all comments at the last step!
For now, those comments help me to plan next test cases.

src/components/gauge/gauge.e2e.ts
MarkMed: Here test two aspects. The first one checks if the default value of the control is assigned if no value is passed. The second one checks if the control takes and assigns the passed value. Do you suggest to merge both tests into only one?

ncardeli: Yes, exactly. DRY also applies to tests 😉

src/components/gauge/gauge.e2e.ts
MarkMed: I forgot to comment on the new fixes! 😅
I removed those comments in the last commit 4e26143

src/components/gauge/gauge.e2e.ts
MarkMed: Check the commit 4e26143, there I merge both tests (checks if the default value of the control is assigned if no value is passed & checks if the control takes and assigns the passed value)

ncardeli (approved):
LGTM!
🚀

PR #185, opened 2020-11-26
Adding Query-Viewer components

snatalevich7:
Query-Viewer for Angular. This is the first push, there are some thing missing yet, also some stuff isn't working.
But for moving forward, to get some validation in case, I need to change some stuff

### DIFF
--- a/src/components/query-viewer-element/query-viewer-element.tsx
+++ b/src/components/query-viewer-element/query-viewer-element.tsx
@@ -1,0 +1,70 @@
+import { Component, Prop } from '@stencil/core';
+// import { Component as GxComponent } from "../common/interfaces";
+
+
+@Component({
+  tag: 'gx-query-viewer-element',
+  shadow: false,
+})
+export class QueryViewerElement {//implements GxComponent {
+
+    @Prop() name: string;
+    @Prop() title: string;
+    @Prop() visible: string;
+    @Prop() type: string;
+    @Prop() axis: string;
+    @Prop() aggregation: string;
+    @Prop() dataField: string;
+    @Prop() filterType: string;
+    @Prop() filterValues: string;
+    @Prop() expandCollapseType: string;
+    @Prop() expandCollapseValues: string;
+    @Prop() axisOrderType: string;
+    @Prop() axisOrderValues: string;
+    @Prop() groupingGroupByYear: boolean;
+    @Prop() groupingYearTitle: string;
+    @Prop() groupingGroupBySemester: boolean;
+    @Prop() groupingSemesterTitle: string;
+    @Prop() groupingGroupByQuarter: boolean;
+    @Prop() groupingQuarterTitle: string;
+    @Prop() groupingGroupByMonth: boolean;
+    @Prop() groupingMonthTitle: string;
+    @Prop() groupingGroupByDayOfWeek: boolean;
+    @Prop() groupingDayOfWeekTitle: string;
+    @Prop() groupingHideValue: boolean;
+    @Prop() raiseItemClick: boolean;
+}
+
+/*
+<gx-query-viewer language='es' charttype='pie'>
+  <gx-query-viewer-elements
+                            name='' 
+                            title='' 
+                            visible='' 
+                            type='' 
+                            axis='' 
+                            aggregation='' 
+                            datafield='' 
+                            filter-type="" 
+                            filter-values=""
+                            expand-collapse-type=""
+                            expand-collapse-values=""
+                            axis-order-type=""
+                            axis-order-values=""
+                            grouping-by-year=""
+                            grouping-year-title=""
+                            grouping-by-semester=""
+                            grouping-*
+                            >
+    <gx-query-viewer-elements-format picture='' subtotals='' style=''>
+      <gx-query-viewer-elements-format-style type="conditional" value='' styleOrClass='' applyToRowOrColumn='' />
+      <gx-query-viewer-elements-format-style type="conditional" value='' styleOrClass='' applyToRowOrColumn='' />
+      
+      <gx-query-viewer-elements-format-style type="format" operator='' value1='' value2='' styleOrClass='' />
+      <gx-query-viewer-elements-format-style type="format" operator='' value1='' value2='' styleOrClass='' />
+    </gx-query-viewer-elements-format>
+      
+  </gx-query-viewer-elements>
+</gx-query-viewer>
+
+*/
--- a/src/components/query-viewer-element/query-viewer-element.tsx
+++ b/src/components/query-viewer-element/query-viewer-element.tsx
@@ -16,19 +16,19 @@ export class QueryViewerElement {
   /**
    * How to show it
    */
-  @Prop() visible: string;
+  @Prop() visible: "Always" | "Yes" | "No" | "Never";
   /**
    * Type of the element
    */
-  @Prop() type: string;
+  @Prop() type: "Axis" | "Datum";
   /**
    * Which axis, row or column
    */
-  @Prop() axis: string;
+  @Prop() axis: "Rows" | "Columns" | "Pages";
   /**
    * Aggregation fucntion
    */
-  @Prop() aggregation: string;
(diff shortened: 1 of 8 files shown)


src/components/query-viewer-element/query-viewer-element.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  @Prop() filterType: string;
+  /**
+   * Filter values comma separated
+   */
+  @Prop() filterValues: string;
+  /**
+   * Expand collapse type
+   */
+  @Prop() expandCollapseType: string;
```
ncardeli, line 47: Looks that this is an enumerated list. We could take advantage of TypeScript to explicitly declare the list of accepted values. For example:

```
@Prop() expandCollapseType = "type1" | "type2" | "type3";
```

src/components/query-viewer-element/query-viewer-element.tsx
```diff
@@ -0,0 +1,108 @@
+import { Component, Prop } from "@stencil/core";
+
+@Component({
+  tag: "gx-query-viewer-element",
+  shadow: false
+})
+export class QueryViewerElement {
```
ncardeli, line 7: Consider initializing some of the properties with default values. Not saying all properties should have a default value, but it's a good practice.

src/components/query-viewer-elements-format-style/query-viewer-elements-format-style.tsx
```diff
@@ -0,0 +1,36 @@
+import { Component, Prop } from "@stencil/core";
+
+@Component({
+  tag: "gx-query-viewer-element-format-style",
```
ncardeli, line 4:
```suggestion
  tag: "gx-query-viewer-format-style",
```

src/components/query-viewer-elements-format-style/query-viewer-elements-format-style.tsx
```diff
@@ -0,0 +1,36 @@
...
+  shadow: false
+})
+export class QueryViewerElementFormatStyle {
```
ncardeli, line 7:
```suggestion
export class QueryViewerFormatStyle {
```

src/components/query-viewer/query-viewer.tsx
```diff
@@ -0,0 +1,332 @@
...
+  @Prop() queryTitle: string;
+
+  private generateForm() {
+    const form = document.createElement("form");
+    form.action = this.baseUrl + this.mapServices[this.env];
+    form.method = "POST";
+    form.target = "query_viewer";
+    form.innerHTML = this.postData();
+
+    document.body.append(form);
```
ncardeli, line 194: I think this could be done inside the `render` method, taking advantage of JSX, instead of manipulating the DOM.
Moreover, it would allow you to simplify all the logic that concatenates strings to create the form.

src/components/query-viewer/query-viewer.tsx
```diff
@@ -0,0 +1,332 @@
...
+      '<input type="hidden" name="Parameters" value="' +
+      this.getParameters() +
+      '" />';
+    return postBody;
+  }
+
+  private getParameters(): string {
+    let parametersValue = "";
+    const parameters = Array.from(
+      document.getElementsByTagName("gx-query-viewer-parameter")
```
ncardeli, line 231: What happens if I programmatically add a new `gx-query-viewer-parameter` child element?

src/components/query-viewer/query-viewer.tsx
```diff
@@ -0,0 +1,332 @@
...
+    return computedStyle.height;
+  }
+
+  render() {
+    return (
+      <div>
+        <iframe
+          name="query_viewer"
+          width={this.getWidth()}
+          height={this.getHeight()}
```
ncardeli, line 327: I think you can avoid calculating sizes by taking advantage of `display: flex; flex: 1;` in all the ancestors of the iframe and `flex: 1;` in the `iframe`.

ncardeli (commented):
Left some initial comments

src/components/query-viewer/query-viewer.tsx
snatalevich7: I can't because I need to send in the post the width & height, for the javascript calculate the size.

src/components/query-viewer/query-viewer.tsx
snatalevich7: It isn't supposed ti happen this.
All parameters should come when it renders or it wouldn't work.

src/components/query-viewer/query-viewer.tsx
snatalevich7: I did it this way, because I needed to submit the form programmatically.

I tried to put in the render method, but I hadn't submit method that way.

src/components/query-viewer/query-viewer.tsx
ncardeli: I'm thinking of the following scenario:

The query viewer is rendered with an initial set of parameters (that were added by the Angular template as `gx-query-viewer-parameter` elements). 
Then, in a GeneXus user event, the set of parameters is changed, which will make the template to render a new set of `gx-query-viewer-parameter`.

With the current approach, we wouldn't be able to render the new parameters. Maybe it's not the case with parameters but what about the other child elements?

src/components/query-viewer/query-viewer.tsx
ncardeli: Ok, understood 👍

src/components/query-viewer/query-viewer.tsx
ncardeli: After rendering it in the `render` method with JSX, you can query the DOM for it using `const form = this.element.querySelector("form");`.

And then you can perform the submit programmatically.

src/components/query-viewer/query-viewer.tsx
snatalevich7: If I understood correct how it will work, it totally make sense if this happens this way.

For example, on Start Event:
receives parameters, CountryName: Spain, CityName: Madrid
Then another event happens and send:
CountryName: Italy, CityName: Rome

We want to replace the previous parameters with the new ones.

And same logic would work for the other child elements

src/components/query-viewer/query-viewer.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+        elementValue += "}";
+      }
+      //TODO add grouping values
+      const formats = Array.from(
+        ax.getElementsByTagName("gx-query-viewer-element-format")
+      );
+
+      formats.forEach(format => {
+        elementValue += ' ,"Format":  {';
```
ggallotti, line 343: Why build a JSON using String concatenation?
Maybe better to just use a json obj?  Less verbose and error prone. 

{
"Name": ax.name, 
"Title": ax.elementTitle,
...
}

src/components/query-viewer/query-viewer.tsx
```diff
@@ -0,0 +1,369 @@
...
+      <input type="hidden" name="Height" value={this.getHeight()} />
+    );
+    postBody.push(
+      <input type="hidden" name="Elements" value={this.getElements()} />
+    );
+    postBody.push(
+      <input type="hidden" name="Parameters" value={this.getParameters()} />
+    );
+
+    return postBody;
```
ncardeli, line 261: Nothing wrong with this approach, but wanted to share with you this more compact and declarative alternative:

```suggestion
    return [
      ...Object
        .keys(this)
        .filter(key => !this.propsNotToPost.includes(key))
        .map(key => <input type="hidden" name={key} value={this[key]} />),
      <input type="hidden" name="Width" value={this.getWidth()} />,
      <input type="hidden" name="Height" value={this.getHeight()} />,
      <input type="hidden" name="Elements" value={this.getElements()} />,
      <input type="hidden" name="Parameters" value={this.getParameters()} />
    ];
```

ncardeli (approved):
Approved! Looks much better now. Just left a couple of comments for you to consider

PR #193, opened 2021-01-19
Fix tab navigation when disabled 

dangmarm:
…avigation

### DIFF
--- a/src/components/tab-caption/tab-caption.tsx
+++ b/src/components/tab-caption/tab-caption.tsx
@@ -3,20 +3,21 @@ import {
   Element,
   Event,
   EventEmitter,
+  Host,
   Prop,
   Watch,
-  h,
-  Host
+  h
 } from "@stencil/core";
 import {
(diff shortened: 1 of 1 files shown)


src/components/tab-caption/tab-caption.tsx
```diff
@@ -124,6 +125,6 @@ export class TabCaption implements GxComponent, DisableableComponent {
 
   private clickHandler(event: UIEvent) {
     event.preventDefault();
-    this.selected = true;
+    this.selected = !this.disabled;
```
ncardeli, line 128: Could a tab caption be initially disabled and selected? If yes, what would happen if the tab caption is clicked?

ncardeli (approved):

src/components/tab-caption/tab-caption.tsx
dangmarm: Visually it is fine, but it is a potential bug.

PR #195, opened 2021-03-16
Fix/control highlight when no action

dangmarm:


src/components/button/button.tsx
```diff
@@ -76,6 +84,11 @@ export class Button
    */
   @Prop() readonly size: "large" | "normal" | "small" = "normal";
 
+  /**
+   * True to highlight control when fire actions.
```
ncardeli, line 88:
```suggestion
   * True to highlight control when an action is fired.
```

ncardeli (approved):
Fantastic!
Just left a small suggestion for a property description.

PR #205, opened 2021-06-01
Fixes gague

ncamera:
- Resizing issues
- Label postioning
- Font resizing
- Responsive implementation

### DIFF
--- a/src/components/gauge/gauge.scss
+++ b/src/components/gauge/gauge.scss
@@ -9,63 +9,67 @@ gx-gauge {
   --center-circle-text-color: rgba(44, 44, 44, 1);
 
   /**
-  * @prop --gauge-border-radius:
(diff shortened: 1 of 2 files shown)


ag-silva (approved):

dangmarm (changes requested):
For the next push make a commit for each fixed issue (Label postioning, Font resizing, etc) instead of a single commit with everything. This way it is easier to understand the intention of each change.

Style values that are fixed, move them to the css file. For example: "margin-top": "1px" (line 169), width / height: 100% (line 213), etc.

In the rotation correction, the problem is due to the width of the marker. I suggest first centering the marker and then doing the rotation (translate (-. 5vw, 0) rotate (... deg)). In this way, the adjust code may not be necessary.

dangmarm (approved):
👍👍

PR #207, opened 2021-06-07
Improved font resizing in circle gauge type and minor fixes.

ncamera:
Main change:
   - In the circle gauge type, solved an issue where the font scaling kept growing even if the container size has reached a maximum size. Now the font (and the value marker) has a maximum size if in the container is defined the attribute "max-height".

Minor changes:
   - Some fixed css were move to the corresponding css file (previous line 211 in .tsx).

   - Removed a div container with the corresponding css selector, due to they don't had any use (previous lines 222 - 228 in .tsx and 159 - 162 in css).

   - In the line gauge type, corrected the position of the value label (previous line 324 in .tsx).

   - Fixed an issue where the line gauge type were not resizing according to the container width (added line 84 in css file).

### DIFF
--- a/src/components/gauge/gauge.scss
+++ b/src/components/gauge/gauge.scss
@@ -71,6 +71,8 @@ gx-gauge {
   */
   --indicator-line-background-color: rgba(44, 44, 44, 1);
 
+  $transitionConfig: 0.3s ease;
+
   @mixin flex-centered {
     display: flex;
     justify-content: center;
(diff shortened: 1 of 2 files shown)


ag-silva (approved):

dangmarm (commented):
In line 213 the 'while' has the condition 'maxSize === undefined' repeated.
Who defines the 'max-height' attribute you are looking for?

dangmarm (approved):
👍

PR #208, opened 2021-06-07
Some tests were modified because they were giving false positives

ncamera:
- Added the property 'show-min-max="true"' to gx-gague definition in the following tests:
       + Second test: "should set the default and the minimum value"
       + Fourth test: "should set the passed maximum value"
       + Fifth test: "should calculate the total ranges values and show the sum as the maximum value"
   Those test were computing either maxValue or minValue to compare with the corresponding span in the gauge.tsx definition. That span only displays if the property showMinMax is set on true. See lines 370 - 383 of gauge.tsx for more information.

   - Changed some selector names in the following tests, due to the recent update in gauge.tsx:
       + Third test: "should set the passed current value"
       + Eighth test: "should properly set a marker (in each gauge type)"
       + Ninth test: "should correctly draw the range (in each gauge type)"

   - Removed some lines and renamed the seventh test "should display the name of range (in each gauge type)". Now is called "should display the name of range (in the line gauge type)", and no longer will check the corresponding ranges in the circle gauge, because the legend implementation was removed from the circle type.

   - Improved the eighth test. Now this test will check if the marker is in the correct position.

### DIFF
--- a/src/components/gauge/gauge.e2e.ts
+++ b/src/components/gauge/gauge.e2e.ts
@@ -18,7 +18,7 @@ describe("gx-gauge", () => {
   /////////////////// Values test ///////////////////////
   it("should set the default and the minimum value", async () => {
     await page.setContent(`
-      <gx-gauge show-value="true"></gx-gauge>
+      <gx-gauge show-value="true" show-min-max="true"></gx-gauge>
     `);
     await page.waitForChanges();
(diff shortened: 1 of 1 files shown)


ag-silva (approved):

dangmarm (approved):
👍

PR #214, opened 2021-06-24
The types of the src and alt properties has been modified, because in…

ncamera:
… the interface 'components.d.ts' they were shown as an enum type


dangmarm (approved):
👍

ag-silva (approved):

PR #217, opened 2021-07-29
Control customization improves

ncamera:
These improvements aim to make more use of the attribute themes defined in GeneXus


dangmarm (changes requested):
In the _base.scss lines were added with !important, we must try to avoid it, more in this case it is a css base. The idea is to pretend that !important does not exist and only use it for exceptions.

dangmarm (commented):
👍

dangmarm (approved):

PR #218, opened 2021-07-30
Checkbox readonly property implementation

ncamera:
In GeneXus, there is a readonly property for the Check Box Attribute, but this property doesn't do anything with the Angular generator.


dangmarm (approved):
👍

PR #224, opened 2021-09-08
`lineClamp` refactoring

ncamera:
- The complexity of the calculations in the `lineClamp` algorithm has been minimized.

- The `lineClamp` algorithm will no longer set the `max-height` property of the components.

- Now, the components will be sized to the maximum available height and the `lineClamp` algorithm will only set the maximum number of lines (`maxLines`) available to display.

- Now, the `lineClamp` algorithm will no longer reset the `maxLines` available to display, instead, the algorithm will always keep the last value of `maxLines` and will update this value if the `content-container` changes his height or the font changes its height.

Those changes has been applied to the gx-edit and gx-textblock controls.

### DIFF
--- a/src/components/common/_base.scss
+++ b/src/components/common/_base.scss
@@ -70,7 +70,6 @@ $gx-xsmall-breakpoint: 768px;
   -webkit-box-orient: vertical;
   -webkit-line-clamp: var(--max-lines);
   overflow: hidden;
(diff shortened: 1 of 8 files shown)


ncamera:
issue:90578

dangmarm (approved):
👍

PR #233, opened 2021-11-10
Fixes for range values in the `date`, `datetime-local` and `time` types

ncamera:
- When the type is `date` or `datetime-local`, we set the maximum date to `9999-12-31`.

- When the type is `date` or `datetime-local`, we set the minimum date to `0001-01-01`.

- When the type is `date`, `datetime-local` or `time`, we set the `step="1"` property to specifies the interval between legal numbers. Without this property, we can not use seconds precision.

- Also, we add `time` type, because we use it in Angular but it was not specified in the `gx-edit` control.

### DIFF
--- a/src/components/edit/edit.tsx
+++ b/src/components/edit/edit.tsx
@@ -163,6 +163,7 @@ export class Edit implements FormComponent, HighlightableComponent {
     | "search"
     | "tel"
     | "text"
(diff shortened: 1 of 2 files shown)


ncamera:
issue: 92149

dangmarm (approved):
👍

PR #235, opened 2021-11-17
gx-tab fixes

ncamera:
- Fix for `--elevation` not working.

- Improves for transition colors in the tab strip.

- Fixes for scroll bars in the tab strip when using Mozilla web browser.

- Implementation of `tabs-distribution` property.


dangmarm (approved):
👍

PR #242, opened 2021-12-10
Animate gauge loading

ncamera:
**Changes we propuse in this PR**:

Add animation to the first load of the gauge childs. In this case, each child will be animated when the `gx-gauge` loads for the first time. 
- The first child will animate with a delay of `0` and an animation duration of `--stroke-animation-duration`

- The second child will animate with a delay of `--stroke-animation-duration` and an animation duration of `--stroke-animation-duration`

- The third child will animate with a delay of `2 * --stroke-animation-duration` and an animation duration of `--stroke-animation-duration`

- And so it goes on.

This creates a feeling of "incremental loading" in the gauge ranges.

The `--stroke-animation-duration: 1s` variable sets the duration of the stroke transition animation the first time the `gx-gauge` is loaded.

issue: 92821

### DIFF
--- a/src/components/gauge/gauge.scss
+++ b/src/components/gauge/gauge.scss
@@ -31,6 +31,13 @@ gx-gauge {
   --stroke-linecap: initial;
 
   /**
+  * @prop --stroke-animation-duration:
+   * Sets the duration of the stroke transition animation the first time the component is loaded.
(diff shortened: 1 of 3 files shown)


dangmarm (approved):
Excellent.

PR #246, opened 2022-01-19
Minor refactoring and improvements to `gx-navbar`

ncamera:
- `issue: 93111` Minor CSS refactoring of the buttons to make them more compatible with the general style of the controls.

- Added color transitions to the buttons.

- `issue: 93040` Added keys to back and toggle buttons to avoid reuse of DOM elements.

- Moved layout config variables like `--gx-navbar-base-z-index`, `--gx-navbar-main-height` and `--gx-navbar-sub-height` to `_layout-styling-config.scss` file.

- `issue: 85664` Improved styling for the bottom navbar.
  - When the width of the screen is less than `$gx-xsmall-breakpoint`, the bottom navbar will be displayed and its style will be the same as the top navbar.

- `issue: 86459` Refactored the way the navbar is used. 
  - Implemented a `position` property that is used to determinate the position of the navbar in the viewport. Bottom navbar is only used to display links when the width of the screen is <= `$gx-xsmall-breakpoint`.
  - The navbar will no longer use a relative container and a fixed content to implement its positioning. Instead, it will use `position: sticky` that serves for this purpose.

- `issue: 94336` Added support for using Shadow DOM in the `gx-navbar` control. Shadow DOM gives us:
  - Faster render slots.
  - An easy way to observer user-driven asynchronous slot changes (eg asynchronous loading of `gx-navbar-item` with `*ngFor`).

- Add theming mixin for `gx-navbar` back and default buttons. In the future, this change will allow us to resolve `issue: 94346`.

- `issue: 94356` The top navbar will no longer be hidden when there are no actions and the toggle button is not visible.

- `issue: 94361` Hide overflow on each line of `gx-navbar`.


dangmarm (commented):
👍

dangmarm (approved):

PR #247, opened 2022-01-28
Housekeeping controls for u8

ncamera:
- `issue: 94512` Fixed `gx-checkbox` and `gx-radio-group` not streching to full width.

- `issue: 94516` Hide `gx-checkbox` caption overflow.

- `issue: 94520` Added support for vertical and horizontal alignment for `gx-checkbox` and `gx-radio-group`.

- `issue: 94515` Controls inside a `gx-table` were not clipped when the `gx-table` has the `border-radius` property set.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -2,10 +2,15 @@
 
 gx-checkbox {
   @include visibility(flex);
-
   @include scss-custom-option-control();
 
   --option-border-radius: 25%;
 
   @include container-margins();
+
+  flex: 1;
+
+  & > [data-readonly] {
+    flex: 1;
+  }
 }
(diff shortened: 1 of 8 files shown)


gmilano (commented):
Just to consider if instead of using constants values like 25% in gx-checkbox, flex: 1, etc
We should use instead variables in order to see in one place all the default values we are using and at the same time we could identify several points of customization for our default components.

ncamera:
> 
Hi!

That's one of our goals when we make some fixes or improvements to the controls, we try to make the default behavior more flexible to change it with a file.
For example, right now we have the files:

- `src\components\common\_layout-styling-config.scss`

- `src\components\common\_common-styling-config.scss`

that centralize styling logic for some controls.

There is still work to be done to make the default behavior more flexible (for example, `flex`, `width` and `height` values) and to expand this mechanism to all the controls.

Great comment. Thanks for the feedback.

dangmarm (approved):

PR #252, opened 2022-03-14
`gx-canvas` refactor and other fixes

ncamera:
  - `issue: 95171` Fix for `gx-canvas-cell` not streching to content height when `auto-grow` is True.

- `issue: 95172` Refactoring for auto-grow implementation in `gx-canvas`.
    - First, we get the minHeight of the `gx-canvas` that is determined by the `gx-canvas-cell`s with `auto-grow = False` (`maxHeight != null`).
     
    - We use a ResizeObserver to monitor size changes in the `gx-canvas-cell` controls with `auto-grow = True` (`max-height == null`). Also, we use another ResizeObserver to monitor size changes in the `gx-canvas` control.

    - Every time the ResizeObserver for the `gx-canvas-cell`s interrupts, we check if there is a cell that has more height than the `gx-canvas` control. Otherwise, we check if the tallest `gx-canvas-cell` decreased its height (in this case, we have to find a new "tallest" `gx-canvas-cell` to correct the height of the `gx-canvas`)

    - If the `gx-canvas` control never had a cell that exceeded its height, then: `this.canvasFixedHeight == null`.

    - When `this.canvasFixedHeight` stops being null, we fix the relative values of `top`, `min-height` and `max-height` in all the `gx-canvas-cell` and we disconnect the ResizeObserver for the `gx-canvas` control.

- `issue: 94810` Hide content overflow in `gx-canvas-cell`.

- `issue: 94783` Hide content overflow in `gx-canvas`.

- `issue: 94733` Removed auto-grow in `width` property of `gx-canvas`.

- `issue: 95071` Added transition colors to `gx-canvas` control.

- `issue: 95124` Add support for `keep-space` and `collapse` visibility in `gx-canvas-cell`.

issue: 95171
issue: 95172
issue: 94810
issue: 94783
issue: 94733
issue: 95071
issue: 95124

_Related issues_.
issue: 92536
issue: 91946
issue: 92699


dangmarm (approved):
👍

PR #259, opened 2022-05-30
Fix gx-grid-horizontal

ncamera:
- Update slidesPerColumn value on `gx-grid-horizontal` swiper when window orientation changes.


dangmarm (approved):
👍

PR #266, opened 2022-07-12
Add icon-image in gx-map-marker

nicoartola:
Adds a property to the marker which allows to modify its icon

### DIFF
--- a/src/components/map-marker/map-marker.tsx
+++ b/src/components/map-marker/map-marker.tsx
@@ -33,7 +33,7 @@ export class MapMarker implements GxComponent {
    * Set the `background-image` property to use it as icon of the marker.
    *
    */
-  @Prop() iconImage = "gx-default-icon";
+  @Prop() iconImageClass = "gx-default-icon";
 
   /**
    * The marker image height.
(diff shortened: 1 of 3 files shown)


ncamera:
Try using `cssClass` as the property name. On most controls, you will find that `cssClass` is the property asociated to the control's class. In this case, the property that you are renaming matches that description:
 - "_The class that the marker will have_"

Also, try updating the readme for this component to keep track of the interface.

By the way, is not the `iconImage` property used in the Angular Generator?

Thanks for the contribution!

nicoartola:
The name of the icon-image property is modified to icon-image-class and also attached the readme

ncamera:
Great! 

One last thing. Readme files are automatically generated by running the `npm run build` command in the root of the project (the same folder that contains the `package.json` file), so you don't have to generate them.

In your case, you will have to: 
 - Run the `npm run build` as it will update the map-marker's readme.

 - After that, in the map-marker's readme there is a tag `<!-- Auto Generated Below -->`, before that tag you should add any comments you fell are necessary to the readme (for example, some of the comments that you added to the `readme-icon-image-class.md` file).

 - And finally, remove the `readme-icon-image-class.md` file as it will not be updated when you run `npm run build`.

 - Commit only the changes to the map-marker's readme and of the removal of the `readme-icon-image-class.md` file. I said this because `npm run build` command makes other changes that are not nedeed in this PR.

Sorry for not telling you before how to update the readme.

nicoartola:
Ready. I remove the **readme-icon-image-class.md** and do the run **npm run build**, thanks for the advice! Here is the commit 31c0c528b8a51422c88c2ea1228c973d5ba799da

PR #268, opened 2022-07-14
Add icon-my-location in gx-map

nicoartola:


src/components/map/map.tsx
```diff
@@ -374,6 +379,27 @@ export class Map implements GxComponent {
...
+        LMarker(coords, {
+          icon: divIcon({
+            className: this.markerClassIcon,
+            iconAnchor: [40, 30],
+            popupAnchor: [0, 10],
+            iconSize: [40, 40],
+            tooltipAnchor: [0, 10]
+          })
+        })
+          .bindPopup("<i>Ubicacion Actual! </i>")
```
ncamera, line 397: Avoid fixed message strings in your code. More even, if they are in spanish. 
If the string can't be bound to a `gx-map`'s property, at least it will be fine if the string is in English.

The best case scenario is to bind this message to a property that can be translated.

nicoartola:
Done. Fixed the recommendations you wrote, thanks for the time! Here is the commit 43f3fa6e4d9ed0b82e62e919ebbf231872f273fd

ncamera (approved):
LGTM 👍

PR #273, opened 2022-08-01
Many fixes

ncamera:
 - Improve support for different use cases of lottie animations.
    - `gx-loading` will now take up space in its container.

   - Changed the way that `gx-loading` is centered (simplified the CSS to center the `gx-loading` on grids).

   - When displayed a slot animation, the `gx-loading` will have a `min-height` of `var(--height)`.

 - Reduce bundle size of grids.

 - `issue: 97413` Fix for `gx-layout`'s targets (asides) being closed on wrong use cases.
   - The targets will only close when the mask is clicked.

 - `issue: 97634` Fix for `gx-image-picker`'s upload button not updating when the image is changed.

 - `issue: 97587` Add `scrollWheelZoom` property to enable/disable mouse wheel zoom.

_Summary of issues_:
issue: 97413
issue: 97634
issue: 97587


dangmarm (approved):
👍

PR #276, opened 2022-08-31
Add support for Inverse Loading in gx-grid-smart-grid. Also, fix resize issues

ncamera:
- Refactoring for `gx-grid-infinite-scroll` implementation.
  - Added support for Inverse Loading (only `direction = "vertical"`).

  - Added support for Shadow DOM.

  - Initials fetchs in `itemCount` Watch will no longer be debounced 100ms.

  - Inifinite scroll position will be `grid-row-start: -(itemCount + 2)` when `position="top"`. This is necessary to ensure that the infinite-scroll will be placed on top of the grid content.

   - An resizeObserver is used to check resizes on the grid content and trigger a callback to check if there is needed to fetch more data in the grid. Also, when `position="top"`, the observer will correct the scroll position to prevent content from jumping.

  - Scroll events are throttled with requestAnimationFrame.

  - The `gxLayoutReference` variable is used when the infinite scroll is attached to the window, since it is needed to compute the position of the infinite scroll relative to the window.

 - `issue: 97977` Add support for Inverse Loading in `gx-grid-smart-css`.
   - `gx-grid-smart-css` supports Inverse Loading when
     - `direction = "vertical"`
     - `itemLayoutMode = "single" | "mbyq"`

   - To change the order of the cells to implement the "Inverse Loading", the `grid-row-start` property is used with the index of the cells. Also, the cell indices are "inverted" (they are negative).

 - `issue: 97239` Fix resize issues in `gx-grid-smart-css`.
   - The `gx-grid-smart-css` will no longer fix its size with the initial available space.

   - The grid cells that have relative size will calculate their corresponing size with the CSS variable `--gx-grid-smart-css-viewport-size`.

   - The CSS variable `--gx-grid-smart-css-viewport-size` will be synced with a resize observer.

 - `issue: 97242` Fix for Snap To Grid not working with drag.

_Summary of issues_:
issue: 97977
issue: 97239
issue: 97242


dangmarm (approved):
👍

PR #277, opened 2022-09-08
Fix for Overflow Behavior implementation

ncamera:
 - `issue: 97976` WA to adjust the canvas height in nested canvases. This WA fix the following scenario:
```  
  <gx-canvas-cell min-height="680px">
    <gx-canvas min-height="90%">
      ...
    </gx-canvas>
  </gx-canvas-cell>
```
 
 - `issue: 98521` Fix for Overflow Behavior implementation.
   - The Overflow Behavior implementation will no longer be in each `gx-table-cell`. Instead (as designed in the documentation) the implementation will be in the `gx-table`.

   - Rename `containerCell` mixin to `align-inner-control-in-cell`.

   - Change `visibility(flex)` mixin to `display: flex` property in the `align-inner-control-in-cell` mixin, since the cells (`gx-table-cell` and `gx-canvas-cell`) do not have `invisible-mode` attribute.

   - Remove `overflowMode` mixin.

   - Added preliminary support for `overflow-style` gx-property in the `gx-table` control.

_Summary of issues_:
issue:97976
issue:98521


dangmarm (approved):
👍

